PR #8, opened 2015-07-20
Make package into a glue plugin

astrofrog:
@nmearl @stscieisenhamer - this makes the package into a proper glue plugin and no longer requires moving the files to the ``~/.glue`` folder. In fact, you should remove any files you copied there in the past or you might end up with two versions of viewers etc.

I don't have immediate access to the datasets you have been testing this with, so could one of you confirm that this works?

Fixes https://github.com/spacetelescope/cube-tools/issues/7


nmearl:
Looks good on my end -- confirmed working. Thanks @astrofrog!

stscieisenhamer:
because i need this now...

PR #37, opened 2017-09-11
Create skeleton application

drdavella:
This is extremely basic (but hopefully moving in the right direction). It's worth mentioning that I believe that if we're testing against stable versions of `astropy` we're going to need to restrict the `pytest` version to v3.1 or earlier, which is the reason for the change to the travis config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,16 +30,17 @@ env:
         - PYTHON_VERSION=3.6
         - NUMPY_VERSION=stable
         - ASTROPY_VERSION=stable
+        - PYTEST_VERSION=3.1
         - MAIN_CMD='python setup.py'
         - SETUP_CMD='test'
         - PIP_DEPENDENCIES=''
         - EVENT_TYPE='pull_request push'
 
-        
+
         # List other runtime dependencies for the package that are available as
         # conda packages here.
         - CONDA_DEPENDENCIES=''
-        
+
         # List other runtime dependencies for the package that are available as
         # pip packages here.
         # - PIP_DEPENDENCIES=''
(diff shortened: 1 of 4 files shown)


nmearl:
The CubeViz package won't be responsible for running Glue itself. Once the package is installed on a system (that must itself already have Glue installed), running Glue proper will give access to the CubeViz plugin. (Unless @astrofrog 's fixed layout changes require directly calling the Glue application?)

Instead, there should be an entry point defined with a `glue.plugins` key that associates the name of the plugin (CubeViz, in our case) with whatever start function is to be called from inside Glue.

Also, side note: I think it might be more \*\~\*\~\*separation of concerns\*\~\*\~\*\~-y to put the entry points inside the `setup.cfg` file instead of directly in the `setup.py` file. (Just personal pref, though!)

drdavella:
Renamed the top-level module containing entry points. Added a stub for glue plugin entry point. Leaving the gui script in place for now as a convenience (and maybe we'll revisit it later). Also, point taken about separation of concerns. It's a good idea but low priority so I haven't changed anything yet.

astrofrog:
@drdavella - is there no way to add the entry points to ``setup.cfg``? If not, then I wonder if we should do a PR to package-template to make that possible, since it would be useful for cases like this (we should avoid changing ``setup.py`` at all if possible to make updates easier)

drdavella:
@astrofrog it's definitely possible but it would require a bit of extra machinery in addition to what's there now. I agree that putting this in `package-template` would be useful, although it could also theoretically go in `astropy_helpers` as a utility function.

nmearl:
You can look at how we did it in [specviz](https://github.com/spacetelescope/specviz/blob/master/setup.cfg#L53); it's not that difficult, since the machinery for entry points defined in the cfg file is already in the [setup.py file](https://github.com/spacetelescope/cubeviz/blob/master/setup.py#L104).

Adding glue plugin recognition to the cfg file is also [pretty straightforward](https://github.com/spacetelescope/specviz/blob/master/setup.py#L113).

PR #38, opened 2017-09-11
Added initial draft of custom CubeViz layout

astrofrog:
**Do not merge!**

**NOTE** the functionality here requires the ``qt-fixed-layout-tab`` branch of glue in my fork for now

This relies on unmerged functionality in the core glue package: https://github.com/glue-viz/glue/pull/1403

For now this doesn't yet add an entry point so one has to manually register this by adding:

```python
from cubeviz import layout
```

to a local ``config.py`` file. The result looks like this for now:

screen shot 2017-09-11 at 18 42 23

This doesn't yet do things like add data automatically to the viewers, or set up any kind of linking, but I plan on experimenting with that next.

We could then combine this with https://github.com/spacetelescope/cubeviz/pull/37 so that when launching ``cubeviz``, it would open glue with this layout by default

### DIFF
--- a/cubeviz/layout.py
+++ b/cubeviz/layout.py
@@ -1,0 +1,68 @@
+from __future__ import print_function, division
+
+from glue.config import qt_fixed_layout_tab
+from qtpy.QtWidgets import QSplitter, QWidget, QVBoxLayout, QHBoxLayout
+from qtpy.QtCore import Qt, QEvent, Signal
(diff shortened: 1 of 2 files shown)


astrofrog:
This now makes it possible to do e.g.:

```
glue JWST-NIRSpec-IFU-GAL2136_CLEAR_PRISM.fits --startu=cubeviz
```

and have the data be pre-loaded into the cubeviz layout

PR #39, opened 2017-09-13
Prototype JWST data cube loader for glue

drdavella:
This PR contains prototype logic for recognizing JWST data cubes (in both ASDF and FITS formats) and loading them as a glue data set. The file recognition uses some pretty simple heuristics that will probably need to be updated/enhanced at some point.

I'm not sure if I'm loading the data correctly at this point, but in any case this will need to be integrated with the work that has been done on glue dashboards in #38.

### DIFF
--- a/cubeviz/utils/loader.py
+++ b/cubeviz/utils/loader.py
@@ -2,21 +2,70 @@
 
 from glue.config import data_factory
 from glue.core import Data
+from astropy.io import fits
+
 import asdf
+from asdf.fits_embed import ASDF_EXTENSION_NAME
 
 
-def is_jwst_data_cube(filename, **kwargs):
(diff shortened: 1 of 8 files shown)


drdavella:
I should have mentioned that these changes provide a 'proper' glue plugin, so once it has been installed, it should be possible to open JWST data cube files using glue.

cubeviz/utils/loader.py
```diff
@@ -0,0 +1,106 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+from glue.config import data_factory
+from glue.core import Data
+from astropy.io import fits
+
+import asdf
```
astrofrog: We should probably add asdf in ``install_requires``

PR #96, opened 2017-11-20
Wavelength unit format

brechmos-stsci:
Cleaned up the view menu
Wavelength text now includes the units (when data loaded)
Wavelength slider value formatted based on wavelength units

image

cc @drdavella @kassin @hcferguson

### DIFF
--- a/cubeviz/layout.py
+++ b/cubeviz/layout.py
@@ -135,9 +135,9 @@ class CubeVizLayout(QtWidgets.QWidget):
         ]
 
         view_menu = self._dict_to_menu(OrderedDict([
-            ('Something', lambda: None),
-            ('Anything', lambda: None),
-            ('Testing', lambda: None)
+            ('RA-DEC', lambda: None),
+            ('RA-Spectral', lambda: None),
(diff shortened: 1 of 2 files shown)


cubeviz/layout.py
```diff
@@ -225,12 +227,27 @@ def _update_slice(self, index):
...
+
+        # Get the wavelength units in order to set the wavelength value's number format
+        self.ui.text_wavelength.setText(self._wavelength_format.format(self._wavelengths[index]))
 
     def _enable_slider(self):
         self.ui.value_slice.setEnabled(True)
         self.ui.value_slice.setMinimum(0)
 
+        # Store the wavelength units and format
+        if self._wavelength_units == None:
```
drdavella: This check might be redundant since I'm not sure whether `self._wavelength_units` will ever be anything other than `None` when `_enable_slider` is called.

brechmos-stsci: good point. I removed it.

brechmos-stsci:
I removed the units check for wavelength number formatting. 3 sigfigs

PR #98, opened 2017-11-20
Move viewer options button to main glue toolbar

astrofrog:
screen shot 2017-11-20 at 16 55 16

### DIFF
--- a/cubeviz/layout.py
+++ b/cubeviz/layout.py
@@ -14,6 +14,7 @@ from glue.core.message import SettingsChangeMessage
 
 from specviz.third_party.glue.data_viewer import SpecVizViewer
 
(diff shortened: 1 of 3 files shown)


brechmos-stsci:
@astrofrog Is there a way to make the button the same style as the "Open Data" button?

nmearl:
I think I lean more toward keeping the way the button is now. Reasoning being it's not a "tool" like the buttons that are part of the toolbar are. It's distinct with an action that's only applicable to the gui.

astrofrog:
I agree with @nmearl, I deliberately made it look different. This is now rebased and ready to merge.

drdavella:
Just rebased these changes on the latest `master`.

PR #101, opened 2017-11-22
Add comments and renamed some variables for more consistency

brechmos-stsci:
This is a work in progress, not to be merged quite yet.


brechmos-stsci:
Small clean up things, @drdavella went through it.  Nothing major here and we can do another round or two later.

cc @kassin @hcferguson

PR #107, opened 2017-11-28
Update requirements in setup.cfg, add conda environment file

drdavella:
This PR adds a conda environment file that provides a single-step solution for creating an environment for alpha testers.

### DIFF
--- a/test-environment.yml
+++ b/test-environment.yml
@@ -22,6 +22,9 @@ dependencies:
   - numpy
   - scipy
   - astropy
+  # This is necessary in some OSX environments to ensure that Python is
+  # installed as a framework
+  - matplotlib
   - pip:
     - git+https://github.com/glue-viz/glue#egg=glue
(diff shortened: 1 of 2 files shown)


test-environment.yml
```diff
@@ -0,0 +1,31 @@
...
+dependencies:
+  - python=3.6
+  - pyqt
+  - numpy
+  - scipy
+  - astropy
+  # This is necessary in some OSX environments to ensure that Python is
+  # installed as a framework
+  - matplotlib
+  - pip:
```
astrofrog: Glue has a bunch of other dependencies which will be installed by pip if not included here. I would recommend looking at https://github.com/glue-viz/glue/blob/master/setup.py#L109 and including as many of those as possible.

drdavella: Do you mean including them in the conda dependency section?

brechmos-stsci:
I ran this yaml file on my machine to create an environment and it worked great.  Merging.

PR #109, opened 2017-11-29
Launch cubeviz layout when loading cube from command line

drdavella:


nmearl:
So, this is a little strange to me: we're embedding Glue GUI layout information in the meta dict of the data objects loaded with cube loaders? I feel like there's some duplication of concern here: we already have a function (the data loader) that recognizes when cube data is being loaded into cubeviz -- can we not simply fire off an event from within the data loader function telling the ui that there is indeed cube data being loaded, instead of passing the actual data object to the front-end? That is, the `receive_event` method of the `HubListener` child doesn't need to have the actual data object to know to configure the ui for cubeviz; a simple `CubeDataAddMessage` fired from within data factory functions should be enough, I'd imagine. This would more cleanly keep the concerns of the ui and the back-end separate.

drdavella:
I agree that the way you propose would be better. The problem as I understand it (and @astrofrog can weigh in) is that the messages from the data loaders will be fired **before** the `cubeviz` layout is ever loaded, so they will be missed. This is the reason that we used this approach to begin with in the last sprint, and this change simply builds off the existing mechanism. It seems like this would require a `glue` infrastructure change in order to support. However, if there's already a message queue, maybe we can make this work.

My suggestion would be to merge this change as-is, and then to revisit this issue later to see if we can clean it up.

nmearl:
That the glue infrastructure *supports* passing data objects doesn't preempt the decoupling of internal vs presented representations of the data we should be wary of. You can imagine that passing data objects is internally useful, but embedding ui information in the internal data objects for use in the front-end that could have otherwise been agnostic is not *really* a practice we should in the habit of.

But you're right in that this is fine for now. It just jumps out to me as something we should err on the side of avoiding when possible.

PR #117, opened 2017-12-01
Arithmetic GUI

brechmos-stsci:
I created a pop-up gui in which one can do "any" type of arithmetic operations on the data components (either pre-existing or newly created).
image

The popup shows the valid data choices that can be used (e.g., FLUX) and will dynamically change if we make calculations. For example, if we do one of the ones below it will show we can also use "FLUXnew". 

For example, one could type the following into the calculation input:
  * FLUXnew = FLUX * 2

  * FLUXnew = FLUX - 1000

Or one could normalize the data between 0 and 255:

   * FLUXnorm = (FLUX - min(FLUX)) / (max(FLUX) - min(FLUX)) * 255

Or one could create a FLUX mask based on the DQ (MASK):

  * FLUXmasked = FLUX * (MASK == 0)

This last one gives:
image

It also reports errors if one types in an incorrect mathematical statement:

image

image

cc: @kassin @hcferguson @robelgeda

### DIFF
--- a/cubeviz/layout.py
+++ b/cubeviz/layout.py
@@ -14,7 +14,7 @@ from specviz.third_party.glue.data_viewer import SpecVizViewer
 from glue.utils.qt import load_ui, get_text
 from glue.external.echo import keep_in_sync
 from glue.utils.qt import get_qapp
-
+from .tools import arithmetic_gui
 
 FLUX = 'FLUX'
 ERROR = 'ERROR'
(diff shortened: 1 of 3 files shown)


brechmos-stsci:
@drdavella and I talked and we think the newly created cube should maybe pop up in the active viewer so there is an indication something happened.  This still needs to happen but maybe doesn't have to hold up this PR.

brechmos-stsci:
@astrofrog @nmearl Could one of you guys take a look at this and try it out real quick? It is reasonably simple but I would like at least one of you to take a look at it before I merge.  @drdavella has gone through it

drdavella:
@brechmos-stsci it will be necessary to add `asteval` to `install_requires` in `setup.cfg` and also to the `pip` dependency section of `test-environment.yml`.

astrofrog:
As mentioned, this functionality already exists in glue itself via the **Add/Edit data components** menu - however, I think the UI in glue needs to be improved. I therefore would like to propose that we get this ready for merging (i.e. Travis passing) but don't invest too much time in improving it further. Once this is merged I can then look into how to improve the built-in functionality in glue to have a UI more similar to the one here, and once that is ready and in a stable release of glue, we can remove the cubeviz one and make it open up the built-in glue one. Does that seem reasonable?

brechmos-stsci:
@astrofrog That seems good to me.  Merge and then re-look at implementation. I checked and the Glue version can do masking etc as I demo'ed in the webex.

What we could do is use the dropdown, layout etc that is in this PR and then just call your code rather than the `ast_eval` package that I used.

@astrofrog, also the CI tests don't mean anything right now, we only started setting them up (I discussed this with @drdavella as well).  So, if you are ok, then could you approve this and we can merge?

PR #119, opened 2017-12-04
Update CI scripts

drdavella:
This PR attempts to update and simplify the CI scripts so that we can begin testing in earnest.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,12 +14,12 @@ sudo: false
 #
 # https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise
 
-addons:
-    apt:
-        packages:
-            - graphviz
-            - texlive-latex-extra
-            - dvipng
+#addons:
+#    apt:
+#        packages:
+#            - graphviz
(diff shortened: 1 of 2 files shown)


drdavella:
There are very strange build failures happening when trying to set up the test environment for `build_docs`. I can't reproduce them locally. It seems to be a conflict between `specutils` and `jinja2`, but I'm not quite sure. I have omitted this test for now, but am leaving a note so it can be revisited later.

.travis.yml
```diff
@@ -30,36 +30,33 @@ env:
         - PYTHON_VERSION=3.6
         - NUMPY_VERSION=stable
         - ASTROPY_VERSION=stable
-        - PYTEST_VERSION=3.1
+        # Pull specviz from github until it is released on PyPi and/or conda
+        - SPECVIZ_GIT='git+https://github.com/spacetelescope/specviz#egg=specviz'
```
nmearl: SpecViz is already available through astroconda, but I do believe you'll have to set the astroconda channel explicitly.

PR #120, opened 2017-12-04
2D moment maps

brechmos-stsci:
The code is implemented to calculate moment maps using spectral-cube. This is modeled on the arithmetic calculations.

This is currently a WIP and should **not** be merged. There is still an outstanding question of how to store the result back in a glue dataset.

### DIFF
--- a/cubeviz/layout.py
+++ b/cubeviz/layout.py
@@ -4,7 +4,9 @@ import os
 from collections import OrderedDict
 
 import numpy as np
+from matplotlib import pyplot as plt
 
+from glue.core.data import Data
 from glue.config import qt_fixed_layout_tab, viewer_tool
 from glue.viewers.common.qt.tool import CheckableTool
 from qtpy import QtWidgets, QtCore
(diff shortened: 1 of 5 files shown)


astrofrog:
@brechmos-stsci - if you want to store the result in a glue dataset, you will need to make a new glue dataset, since glue enforces that the shape/dimensions have to match between components of a dataset. There's no reason we can't have a second dataset though?

brechmos-stsci:
This grew a little bit more than just the moment maps, in order to accomodate the moment maps.  So, this PR has:

1. Moment maps calculated using spectral-cube

2. Currently, they are displayed as an overlay on each of the 3 views:
image

3. We added in the ability to show/hide any of the overlays, select a colormap for the overlay and to interactively change the transparency of the overlay.
image

4. As an aside, as part of this we made some changes to the upper toolbar in order to *group* the similar functionality (e.g., overlay functionality, or slice functionality).  
image

brechmos-stsci:
@astrofrog @nmearl  Could one of you guys take a look at this PR and approve if you have no major qualms? 

cc @kassin @hcferguson

drdavella:
@brechmos-stsci, @astrofrog, @nearl, I have another branch with the exact same commit history but it has been rebased on `master` instead of merging `master` back into the feature branch. This might keep the history a bit cleaner, but I'm not sure if it's a big deal. Just thought you might be interested before merging.

PR #121, opened 2017-12-05
Split out custom image viewer into separate file

astrofrog:
To keep files from growing too much, I think it'd make sense to split out the custom image viewer stuff into its own file.


drdavella:
@brechmos-stsci and I are working on a refactoring of the `layout.py`, so we may just cherry-pick this onto our work since it seems like a good idea.

brechmos-stsci:
@drdavella Is this ticket closeable?  ie have you/we cherry picked what we need out of it?

PR #123, opened 2017-12-05
WIP: Change aspect ratio depending on visible axes

astrofrog:
This makes it so that the pixels are square if looking at the spatial axes, but not for other combinations of axes (for example for declination-spectral, there's no reason the pixels should be square).

However, note that this has some issues with the syncing between viewers, due to the way Matplotlib enforces the square aspect ratio. I'm going to try and find a solution to this, but in the mean time we should probably refrain from merging this.


astrofrog:
I'm going to close this since I don't think we really support not having the slicing be over the spectral axis right now anyway.

PR #125, opened 2017-12-05
WIP: Use combo box helpers

astrofrog:
This will cause the combo boxes to update automatically if new components are added This still requires some work - in particular ``add_new_data_component`` shouldn't add items manually to the combo anymore, etc. 

@drdavella - feel free to use this commit in another PR or add commits here

### DIFF
--- a/cubeviz/layout.py
+++ b/cubeviz/layout.py
@@ -12,7 +12,9 @@ from qtpy.QtWidgets import QMenu, QAction
 from glue.viewers.image.qt import ImageViewer
 from specviz.third_party.glue.data_viewer import SpecVizViewer
 from glue.utils.qt import load_ui, get_text
-from glue.external.echo import keep_in_sync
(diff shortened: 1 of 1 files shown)


drdavella:
Thanks @astrofrog. I'll try to cherry-pick this commit into some other UI improvements I'm doing. My main concern about this approach is does it limit us to only ever displaying components from a single data collection in a particular viewer? It sounds like you're making some changes to the kinds of data that can be displayed in a single viewer, but will those changes be reflected in these combo boxes as well?

drdavella:
Also just wanted to mention that these new combo boxes do not appear to be affected when the attributes are modified from the `Plot Layers` control box under `Glue Viewer Options`. Is there any way to make the interaction work in that direction as well?

drdavella:
Closing since this has been subsumed by #130.

PR #130, opened 2017-12-08
WIP: More layout and UI improvements

drdavella:
This PR adds a few improvements to the user interface:
* it allows individual viewer toolbars to be hidden
* it allows the axes in the viewers to be hidden so the image can expand to fill the entire viewer
* it restores and updates the functionality of the single image viewer

Also, it integrates the changes that @astrofrog made in #125, so that PR can be closed.

**NOTE**: It is not yet ready for merge since I still need to work out some of the interface logic since we need to account for the interaction between the split image layout and the single image layout.

### DIFF
--- a/cubeviz/controls/overlay.py
+++ b/cubeviz/controls/overlay.py
@@ -10,7 +10,7 @@ class OverlayController:
 
     def __init__(self, cubeviz_layout):
         self._cv_layout = cubeviz_layout
-        self._cubes = cubeviz_layout.cubes
+        self._cube_views = cubeviz_layout.cube_views
         ui = cubeviz_layout.ui
 
         self._overlays = Data('Overlays')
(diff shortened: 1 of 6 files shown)


drdavella:
**UPDATE**: This is nearly ready for merge. However, the axes hiding behavior is not working quite as expected in all circumstances. I think this is because we are not currently accounting for some underlying `glue` mechanism that redraws the `mpl` viewers. I have reached out to @astrofrog about this and hopefully he can help me figure out the right solution.

drdavella:
Note to self: this should be split into two separate PRs: one with the changes for single viewer support, since this can be merged now; and another with the feature for toggling axes, since this might need some help from `glue` after all.

drdavella:
This PR has been superseded by #149 and #150.

Issue #116, opened 2017-12-01
Dashboard whitespace


hcferguson:
This is a sort of generic visualization philosophy issue for me and you can feel free to ignore it.

We are now using a very large fraction of the screen real-estate for labels, menus and menu bars, status bars, and whitespace. That may be what most people like for the default view and it certainly has the advantage that you can see all the menu options and coordinates, etc. But it means that the views of the actual data are rather small. 

I can think of several was that one might like to toggle the view (e.g. with a hotkey) to get rid of distracting info. 

- Toggle from the side-by-side tiled view of the data, err & dq to the stacked view. In the default layout this doesn't really help if the windows stay the same size. So I guess we would need to think about whether the alternate dashboard view has a completely different layout (e.g. horizontal with the spatial view on the left and spectral view on the right). I can't really visualize whether this would be very useful. Maybe someone has already played with this?

- Toggle the current window to fill the canvas. For the spatial view this could be quite handy. For the spectral view, I could imagine wanting it to change the aspect ratio so I can get more sense of the shapes of spectral features than I can get from the squished horizontal view in the default. Of course one can just drag the handle on this window and do that sort of thing, so making this a toggle isn't a particularly high priority.

- Toggle on and off the labeling on the spatial views. I actually think a better default for these windows would be not to label the axes and use the real-estate to display the pixels. Especially now that we have the cursor readback. Add the option to switch the cursor readback between pixels and RA,Dec and most of the time you don't really need the labels. (You do want to be able to have them for making hardcopies though; but for example in ds9 that's really the only time I ever turn on coordinate labeling).

- Toggle on and off the menu bars with a hotkey. Default to having them on; advanced users will eventually figure out how to toggle them off.

brechmos-stsci:
1, 2, 4 are basically completed by @drdavella and are in a PR. 

Point 3 will have to be discussed with @hcferguson to determine importance etc.

brechmos-stsci:
@drdavella I believe this is done, correct?  If so, please close

drdavella:
We haven't implemented toggling the current view to fill the entire canvas yet. I'm not sure how important that is, though.

brechmos-stsci:
Given the other more science aspects I think that will be a future improvement (i.e., after the next two sprints).

drdavella:
All of these suggestions have been implemented, with the exception of toggling the current window to fill the canvas. Closing this issue, and we will track the remaining feature as part of our effort to make the layout more flexible.

PR #149, opened 2018-01-03
Restore single image viewer functionality

drdavella:
PR #130 is being broken up into two separate PRs. This new PR contains only the single image viewer functionality, along with some more general layout improvements. This is also the portion of the original PR that contains the changes introduced in #125.

This partially addresses issue #116.

### DIFF
--- a/cubeviz/controls/overlay.py
+++ b/cubeviz/controls/overlay.py
@@ -10,7 +10,7 @@ class OverlayController:
 
     def __init__(self, cubeviz_layout):
         self._cv_layout = cubeviz_layout
-        self._cubes = cubeviz_layout.cubes
+        self._cube_views = cubeviz_layout.cube_views
         ui = cubeviz_layout.ui
 
         self._overlays = Data('Overlays')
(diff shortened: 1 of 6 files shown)


kassin:
Apologies if we've been through this before (I forget), but would it be possible to allow the user to resize the viewers in the single image viewer in the vertical direction and even allow the image and spectrum viewers to overlap?

brechmos-stsci:
@kassin I don't remember a discussion about that. 

I think to resize the viewers the best is to just resize the whole window.  I suspect the way it is implemented it is not possible to resize one viewer (image or spec) in the single viewer. 

What would be the use case for this?

kassin:
In playing around with CubeViz I kept wanting to resize either the image or the spectrum viewer to encompass my full screen (and I was working with a large screen).  If users can resize windows and allow them to overlap, they can create custom "dashboards" to suit their particular workflow.  That said, I imagine this is not something that is easily done?  @astrofrog do you have thoughts on this?

brechmos-stsci:
@kassin Maybe best to create a new issue for this and maybe tag it with the "feature-request" label and we can discuss there.

kassin:
window resizing is now #157

PR #150, opened 2018-01-03
Toggle whitespace

drdavella:
The changes in this PR were originally part of PR #130, which has now been broken into separate PRs representing different functionality. The work in this PR is currently based on #149, but should eventually be rebased onto `master` once that PR is merged.

This PR provides the following functionality:
* Hiding viewer toolbars
* Hiding viewer axes

However, the axes hiding behavior is not working quite as expected in all circumstances. I think this is because we are not currently accounting for some underlying `glue` mechanism that redraws the `mpl` viewers. I have reached out to @astrofrog about this and hopefully he can help me figure out the right solution.

This PR partially addresses #116.

### DIFF
--- a/cubeviz/layout.py
+++ b/cubeviz/layout.py
@@ -132,6 +132,9 @@ class CubeVizLayout(QtWidgets.QWidget):
         # This tracks the current positions of cube viewer axes when they are hidden
         self._viewer_axes_positions = []
 
+        # Indicates whether cube viewer toolbars are currently visible or not
+        self._toolbars_visible = True
+
         self._slice_controller = SliceController(self)
         self._overlay_controller = OverlayController(self)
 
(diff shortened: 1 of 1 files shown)


astrofrog:
@drdavella - I found the issue with the axes in glue getting reset - the issue is that in glue we use a helper function called ``freeze_margins`` to make sure that the margins always stay the same in absolute terms rather than relative terms:

https://github.com/glue-viz/glue/blob/master/glue/utils/matplotlib.py#L254

Instead of using ``set_position``, do:

```python
axes.resizer.margins = [0, 0, 0, 0]
axes.resizer.on_resize(None)
```

In case some viewers don't use ``freeze_margins``, you could always check if ``resizer`` exists, and if so do the above, otherwise just do what you are doing currently.

drdavella:
@astrofrog thanks! I had to play around with it a little bit, but I think I finally got things working.

@brechmos-stsci this should be working now, and should be ready for merge. If you don't mind doing a quick review/test, that would be great.

PR #160, opened 2018-01-12
Re-factor the data factory

brechmos-stsci:
The previous data factory had if-statements in the code in order to rank the data type based on the input files. This would require re-working the code each time a new ifu type was introduced.  This is now re-worked in order to read in YAML files that define configurations for each image type.  An input data file is then ranked based on the matching YAML file configurations.

~~There is still some testing I need to do, so **don't merge yet**.~~


brechmos-stsci:
@drdavella could you please look through this and test it on data files that you *know* work.  If you want to try other YAML files, copy out of  the main area and put into a directory, then you can either include the directory in an environment variable CUBEVIZ_DATA_CONFIGS or as a parameter on the command line `--data-configs <directory>`.

cubeviz/data_factories/__init__.py
```diff
@@ -2,3 +2,287 @@
...
+from ..layout import FLUX, ERROR, MASK
+
+from os.path import basename, splitext
+import yaml
+import os
+import glob
+import logging
+from astropy.io import fits
+import numpy as np
+from glue.config import data_factory
```
drdavella: A bit nit-picky, but [according to PEP8](https://www.python.org/dev/peps/pep-0008/#imports):
> Imports should be grouped in the following order:
> * standard library imports
> * related third party imports
> * local application/library specific imports
> 
> You should put a blank line between each group of imports.

cubeviz/data_factories/__init__.py
```diff
@@ -2,3 +2,287 @@
...
+            mask_data = hdulist[self._data['DQ']].data
+        else:
+            mask_data = np.empty(flux_data.shape)
+
+        label = "{}: {}".format(self._name, splitext(basename(data_filename))[0])
+
+        data = Data(label=label)
+
+        data.coords = coordinates_from_header(flux.header)
+        data.meta[CUBEVIZ_LAYOUT] = self._name
```
drdavella: Admittedly this is missing from the original code, but it would be useful to have a comment here about how this attribute is used to indicate to the `cubeviz` layout that this is a cubeviz-specific data component.

cubeviz/data_factories/__init__.py
```diff
@@ -2,3 +2,287 @@
...
+
+
+        # Now call the internal processing.
+        matches = self._process('all', self._configuration['all'])
+
+        if matches:
+            logger.debug('{} matches {}'.format(self._config_file, filename))
+            return True
+        else:
+            return False
```
drdavella: This could be simplified slightly with the following:
```python 
if matches:
    logger.debug('{} matches {}'.format(self._config_file, filename))

return matches
```

cubeviz/data_factories/__init__.py
```diff
@@ -2,3 +2,287 @@
...
+        if 'all' == key:
+            return self._all(conditional)
+        elif 'any' == key:
+            return self._any(conditional)
+        elif 'equal' == key:
+            return self._equal(conditional)
+        elif 'startswith' == key:
+            return self._startswith(conditional)
+        elif 'extension_name' == key:
+            return self._exists(conditional)
```
drdavella: Poor Python without a `switch` statement. Sometimes I wonder if this kind of code would be improved by a `dict`, but it's not really important.

cubeviz/data_factories/configurations/jwst-fits.yaml
```diff
@@ -0,0 +1,35 @@
+# JWST FITS cube
+name: 'jwst-fits-cube'
+type: 'JWST-FITS-CUBE'
+
+# Highest priority data configuration file, that matches, is selected.
+priority: 1200
+
+# All must match and they must have SCI, ERR, DQ extensions
+match:
```
drdavella: Eventually we will need to figure out a way to handle other data formats such as `ASDF` and possibly `HDF5`.

cubeviz/cubeviz.py
```diff
@@ -1,23 +1,98 @@
...
+    parser.add_argument('data_files', nargs=argparse.REMAINDER)
+    args = parser.parse_known_args(argv[1:])
+
+    # Store the args for each ' --data-configs' found on the commandline
+    global global_data_configuration
+    global_data_configuration['data_configs'] = args[0].data_configs if args[0].data_configs else []
+    global_data_configuration['data_configs_show'] = args[0].data_configs_show
+
+    import glue
+    from glue.utils.qt import get_qapp
```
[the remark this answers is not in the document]
brechmos-stsci: moved

cubeviz/cubeviz.py
```diff
@@ -1,23 +1,98 @@
...
+
+    hub = None
+
+    # Show the splash screen for 1 second
+    from qtpy.QtCore import QTimer
+    timer = QTimer()
+    timer.setInterval(1000)
+    timer.setSingleShot(True)
+    timer.timeout.connect(splash.close)
+    timer.start()
```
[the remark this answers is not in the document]
brechmos-stsci: I am going to leave this one for now as it is the way glue has it and I don't want to change the look of the code too much from glue

cubeviz/cubeviz.py
```diff
@@ -1,23 +1,98 @@
...
+
+    # Process glue file if passed in (was an argument to start_glue()
+    gluefile = None
+    if gluefile is not None:
```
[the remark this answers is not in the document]
brechmos-stsci: removed

cubeviz/cubeviz.py
```diff
@@ -1,23 +1,98 @@
...
+        app = restore_session(gluefile)
+        return app.start()
+
+    # Process config file if passed in (was an argument to start_glue()
+    config = None
+    if config is not None:
```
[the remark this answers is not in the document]
brechmos-stsci: removed

cubeviz/cubeviz.py
```diff
@@ -1,23 +1,98 @@
...
-            break
-    else:
-        argv.append('--startup=cubeviz')
+    # Load the data files.
+    if datafiles:
+        datasets = load_data_files(datafiles)
+        ga.add_datasets(data_collection, datasets, auto_merge=False)
 
-    sys.exit(glue_main(argv))
+    return ga.start(maximized=True)
```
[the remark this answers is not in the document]
brechmos-stsci: added sys.exit

cubeviz/data_factories/__init__.py
```diff
@@ -2,3 +2,287 @@
 from .jwst import read_jwst_data_cube
 from .kmos import read_kmos_data_cube
 from .manga import read_manga_data_cube
```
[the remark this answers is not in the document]
brechmos-stsci: removed

cubeviz/data_factories/__init__.py
brechmos-stsci: switched around

cubeviz/data_factories/__init__.py
```diff
@@ -2,3 +2,287 @@
...
+
+        :param data_filename:
+        :return:
+        """
+        hdulist = fits.open(data_filename)
+
+        try:
+            flux_index = int(self._data['FLUX'])
+        except:
+            flux_index = self._data['FLUX']
```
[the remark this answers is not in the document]
brechmos-stsci: Going to see if it is an integer, if not then use the string value.  Some files are indexed by 0 others by 'FLUX', for example.

cubeviz/data_factories/__init__.py
```diff
@@ -2,3 +2,287 @@
...
+        flux = hdulist[flux_index]
+        flux_data = flux.data
+
+        try:
+            error_index = int(self._data['ERROR'])
+        except:
+            error_index = self._data['ERROR']
+        var_data = hdulist[error_index].data
+
+        if not self._data['DQ'] == 'None':
```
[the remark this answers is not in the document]
brechmos-stsci: The file contains the string None which is why I compared to the quoted string None.

cubeviz/data_factories/__init__.py
brechmos-stsci: I added a comment in there.

cubeviz/data_factories/__init__.py
brechmos-stsci: done

cubeviz/data_factories/__init__.py
```diff
@@ -2,3 +2,287 @@
...
+            name = cfg['name']
+
+            try:
+                priority = int(cfg.get('priority', 0))
+            except:
+                priority = 0
+
+            dc = DataConfiguration(config_file)
+            wrapper = data_factory(name, dc.matches, priority=priority)
+            wrapper(dc.load_data)
```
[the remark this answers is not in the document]
brechmos-stsci: I added a comment.

cubeviz/data_factories/configurations/jwst-fits.yaml
brechmos-stsci: What is ASDF?

brechmos-stsci: bhahahahahahhahaa

PR #161, opened 2018-01-12
Add meaningful unit tests

drdavella:
These tests attempt to validate basic UI functionality. Tools such as the arithmetic GUI, smoothing GUI, etc. are not tested here. Hopefully these tests can serve as a template for those components.

Also, there is really no testing of the `specviz` UI at this point, although that can be added later.

Once this PR is accepted, it means that people merging subsequent PRs will actually need to pay attention to the CI testing status.

This closes #50 and also closes #53. It also now closes #162.

PR #165, opened 2018-01-18
Fix pinned pytest version in default_install_requires

drdavella:
This fixes a minor oversight. Pytest no longer needs to be pinned to a specific version. This was fixed in #161, but it was not reflected in the default fallback versions of packages.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -122,7 +122,7 @@ for root, dirs, files in os.walk(PACKAGENAME):
 package_info['package_data'][PACKAGENAME].extend(c_files)
 
 # I'm not sure whether it makes sense to duplicate this here or not
-default_install_requires = 'astropy asdf pytest==3.1'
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -122,7 +122,7 @@
 package_info['package_data'][PACKAGENAME].extend(c_files)
 
 # I'm not sure whether it makes sense to duplicate this here or not
-default_install_requires = 'astropy asdf pytest==3.1'
+default_install_requires = 'astropy asdf pytest>=3.1'
```
astrofrog: Regarding the comment above, maybe the answer is 'no' and we could simply do ``metadata['install_requires']`` below?

drdavella: I'm not opposed in principle, but right now every other field that we get from the `metadata` dict has a default value. Should we git rid of all of those as well?

Issue #148, opened 2018-01-03
Make sure display layout is reasonable on other platforms


drdavella:
Make sure fonts and layout don't look terrible on various Linux distros, and possibly Windows. Make sure to check on both larger and smaller screens.

brechmos-stsci:
@drdavella Given #168 and the recent work you have done on your linux system can we close this for now and open up a new issue if there are specific things brought to us?

PR #168, opened 2018-01-18
Remove hard-coded font sizes

astrofrog:
Hard-coded absolute font sizes should not be use in ui files (nor in Python code) as the default font size is platform dependent, so hard-coded absolute font sizes look weird on some platforms. This adds a check to Travis to make sure we don't reintroduce any. The proper way to change font sizes is to do the following in the Python code - using the example of a QLabel object:

```python
label = self.ui.label_title
font = label.font()
font.setPointSize(font.pointSize() * 1.5)
label.setFont(font)
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -133,6 +133,8 @@ install:
 
 script:
    - $MAIN_CMD $SETUP_CMD
+   - find glue -name "*.ui" -exec grep "pointsize" {} \; >& font.log
+   - test ! -s font.log
 
 after_success:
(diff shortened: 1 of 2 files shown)


drdavella:
Failing unit tests will be fixed by #172, so this PR will need to be rebased.

I tested this out on my Linux laptop, and this makes the general UI look **dramatically** better. In fact, I think this mostly, or maybe entirely, addresses the concerns of #148.

PR #169, opened 2018-01-18
Make proper glue dataset to hold moment maps

astrofrog:
The idea here is to store the moment maps in a separate glue dataset rather than just internally in the cubeviz layout. There are several benefits to doing this:

* We can use the glue machinery to deal with showing the overlay rather than having to maintain our own custom way of doing it (since the glue image viewers allow multiple layers to be shown with transparency)

* We can show *just* a moment map in one of the viewers, we don't have to show it as an overlay

* We can also open a new tab and use the moment maps as a full glue dataset.

### DIFF
--- a/cubeviz/image_viewer.py
+++ b/cubeviz/image_viewer.py
@@ -708,7 +708,7 @@ class CubevizImageViewer(ImageViewer):
                             string = string + " " + self._coords_format_function(ra, dec)
                 # Pixel Value:
                 v = arr[y][x]
-                string = "{0:1.4f} {1} ".format(v, self.component_unit_label) + string
+                string = "{0:.3e} {1} ".format(v, self.component_unit_label) + string
         # Add a gap to string and add to viewer.
         string += " "
         self._dont_update_status = True
--- a/cubeviz/image_viewer.py
+++ b/cubeviz/image_viewer.py
@@ -681,6 +681,11 @@ class CubevizImageViewer(ImageViewer):
         if len(self.state.layers) > 0:
             # Get array arr that contains the image values
             # Default layer is layer at index 0.
+            for layer in self.state.layers:
+                if layer.layer is self.state.reference_data:
+                    arr = layer.get_sliced_data()
+            else:
+                raise Exception("Couldn't find layer corresponding to reference data")
             arr = self.state.layers[0].get_sliced_data()
             if 0 <= y < arr.shape[0] and 0 <= x < arr.shape[1]:
                 # if x and y are in bounds. Note: x and y are swapped in array.
--- a/cubeviz/image_viewer.py
+++ b/cubeviz/image_viewer.py
@@ -324,7 +324,8 @@ class CubevizImageViewer(ImageViewer):
 
     def get_contour_array(self):
         if self.contour_component is None:
-            arr = self.state.layers[0].get_sliced_data()
+            layer_artist = self.first_visible_layer()
+            arr = layer_artist.state.get_sliced_data()
         else:
             data = self.state.layers_data[0]
             arr = data[self.contour_component][self.slice_index]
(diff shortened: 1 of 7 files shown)


brechmos-stsci:
@astrofrog This is marked as a WIP, is this still being worked on?  I sort of thought you had said that this is now working (or was that a Glue functionality thing)?

brechmos-stsci:
@astrofrog Is this close to being able to be merged (aside from the conflicts and failing tests?)  I will be (hopefully) merging the cube collapse will also create 2D images from the 3D data sets.  The ability to process the 2D images will be super helpful.

astrofrog:
(I'm working on this locally, will push soon)

astrofrog:
@brechmos-stsci - this is ready for review. What this does is that any dataset loaded into cubeviz will now have a ``contained_2d`` dataset created which has the same filename but with ``[2d]`` suffix. This is created as-needed, when moment maps are created. The idea is that this single dataset then contains any collapsed 2d map for a given cube. So this could then be used for the 'collapse cube' functionality too (not just moments).

brechmos-stsci:
Cool, but a couple of things... 

1. I think it would be useful to keep the RA/DEC axes so it is easy for people to compare locations.

2. The intensity value is displaying as "0" when I hover over spaxels in the image (bottom right of the image)

image

3. There seems to be a bug. When I create a moment map, turned off the overlay and displayed the moment map in the middle viewer, I don't seem to be able to change the spaxel window/level as the toolbar thing says "Cannot visualize this layer..." (left side)  I think I should be able to change the spaxel intensities, no?

image

4. The contour does not seem to work on the 2D component either.  I would think it should be able to work as it is just a representation of the 2D data being displayed.

Thoughts on those?

astrofrog:
@brechmos-stsci - I haven't yet pushed the latest changes here because I'm waiting on some bug fixes to be merged into glue, so no need to review yet, but here are some comments regarding your questions

1. Fixed

2. This turned out to be a bug in cubeviz which was that the value was shown using the format ``1.4f`` - which means any value < 0.0001 would not be shown. I've switched to scientific notation.

3. Fixed

4. Fixed, somewhat - basically the contour will work if you select to show the contours on the current component, but what doesn't work yet is the ability to overlay e.g. the moment 0 contour on top of the cube components. Given the size of this PR, I'd like to delay this to a future PR (which I can work on) since this isn't a regression, but a missing feature.

I've made some further changes in this PR which have to do with the fact that the previous code seemed to always assume that the relevant layer in any viewer was the first one - this assumption breaks when a second dataset is added to the viewers. Similarly, the combo boxes for the viewers now have separators that give the name of the dataset, so any assumptions about a relation between component and index in the combo was broken. Finally, there were a number of places that relied on component labels, when the more robust way to do it is to pass around ComponentID instances. I've updated the code to make it more robust - for example ``change_viewer_component`` now takes a component ID instead of an index. I'll go through the code after next week to find any other places that can be generalized along these lines, but this PR deals with the cases that matter here to preserve functionality.

Finally, I fixed a bug which was that after the smoothing preview was stopped, the viewer combo would previously be reset to the first component even if that wasn't the originally selected one.

astrofrog:
I've now also extended this to the cube collapsing so that collapsed cubes can be used in the arithmetic too - e.g. one line minus another line, or minus the continuum.

brechmos-stsci:
Still one thing I don't see is the ability to change the window and level when the 2D component is displayed.  Currently it shows "Cannot visualize this layer:..." (bottom left) Could this be tweaked as I can see it would be important to be able to change window/level on the 2D components just like the 3D components:
image

astrofrog:
@brechmos-stsci - huh, this should work already, did you update to the very latest developer version of glue-core?

astrofrog:
This is what it looks like for me: notice that the active plot layer selected in the bottom left changes as I change the component. This requires the very latest version of glue.

layers

brechmos-stsci:
ok, I will try again, maybe it was the fact that glue wasn't update to the latest.

PR #172, opened 2018-01-18
Make sure that data factory configuration files are installed

drdavella:
This should fix unit tests that have been failing since #160 was merged.


brechmos-stsci:
All tests passed except Mac but we are going to merge anyway.

PR #176, opened 2018-01-22
Display accurate development version in main window title

drdavella:
image


brechmos-stsci:
This was needed so we have a reference point in discussions with users and other developers. For example, to be able to say a certain dataset loaded I needed to be able to reference the "current" version.

brechmos-stsci:
All required tests are good (mac one still hanging).

PR #181, opened 2018-01-23
Read the docs

brechmos-stsci:
There were a couple small changes that had to happen in order to get the builds to work on the Read the Docs.  With the two changes below we now have a page http://cubeviz.readthedocs.io/en/latest/:

image

### DIFF
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -1,5 +1,6 @@
 conda:
-  file: .rtd-environment.yml
+  file: test-environment.yml
 
 python:
    setup_py_install: true
+   version: 3.6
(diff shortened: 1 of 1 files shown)


readthedocs.yml
```diff
@@ -1,5 +1,6 @@
 conda:
-  file: .rtd-environment.yml
+  file: test-environment.yml
 
 python:
    setup_py_install: true
+   version: 3.5
```
drdavella: Any reason to tie this to 3.5 rather than 3.6? Does it need to be tied to a version at all?

brechmos-stsci: 3.6 failed in read the docs. Let's leave at 3.5 for now and we can look at it in the future.

PR #199, opened 2018-01-26
Wavelength Unit conversion

javerbukh:
Wavelengths are being converted and the Wavelength label converts as well. Next step is to get wavelengths to update within layout.py.


brechmos-stsci:
There appears to be one bug but I believe it is in specviz.  When the units change the x-axis label changes the units but the scientific notation doesn't appear to change until the specviz plot is updated. I'll add an issue to specviz if confirmed.

PR #207, opened 2018-02-07
Slider hotkey

javerbukh:
Uses the left and right arrow keys to move the slider. Tests have also been written that show the hotkeys work. However, when actually running cubeviz and using the keys, they move by 2 instead of by 1. This might be a continuation of the previously known bug where events fire multiple times for one press


drdavella:
I'll try to take a look at this tomorrow. In the meantime, there are a lot of test failures here. Most of them come from the fact that we are testing against stable versions of `glue`, so they don't include the latest updates to keyboard shortcuts. However, this test failure needs more investigation since it's against the latest development version of `glue`: https://travis-ci.org/spacetelescope/cubeviz/jobs/338555630

drdavella:
@javerbukh is this PR still relevant? If so it will need to be rebased so that the tests can run against the latest development version of glue (see #219).

javerbukh:
It is, I'll do that now

Issue #190, opened 2018-01-25
Open Dialog Box - More Info


kassin:
From @bethanjames:

"It wasn't clear where the data were going once you've loaded them in. Perhaps include a message/note within the ‘open’ dialogue box to tell that the data will be found in Data Collection on the left hand panel once read in."

I agree, this would be a good idea to add.

@brechmos-stsci

brechmos-stsci:
The dataset in question `/astro/3/jwst_da_sprint_testdata/IFU_datacubes/MUSE/MUSE_CosmicHorseshoe_bjames.fits` was not readable by cubeviz but cubeviz was not showing that there was an error in reading.

brechmos-stsci:
this should now be taken care of with PR #210

PR #210, opened 2018-02-09
Bad data popup

brechmos-stsci:
This will show a popup that says there is bad data if it is an unrecognized IFU datafile, OR if there is a problem with the data file and the data loader can't parse it.

One data file that fails will now popup:
image
rather than show a long traceback on the command line.

Another data file (a MUSE format that isn't recognized) now shows:
image
rather than bringing up cubeviz with no data  in the image viewers.

Addresses #190

### DIFF
--- a/cubeviz/cubeviz.py
+++ b/cubeviz/cubeviz.py
@@ -6,6 +6,11 @@ from glue.app.qt import GlueApplication
 from glue.main import restore_session, get_splash, load_data_files, load_plugins
 from qtpy.QtCore import QTimer
 
+try:
+    from glue.utils.qt.decorators import die_on_error
+except ImportError:
+    from glue.utils.decorators import die_on_error
+
 from .version import version as cubeviz_version
 
 # Global variable to store the data configuration directories/files
@@ -23,6 +28,7 @@ def setup():
     from . import layout  # noqa
     from . import startup  # noqa
 
+@die_on_error("Error starting up Cubeviz")
 def main(argv=sys.argv):
     """
     The majority of the code in this function was taken from start_glue() in main.py after a discussion with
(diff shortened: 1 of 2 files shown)


brechmos-stsci:
@astrofrog Is there a better way to do de-register members of the data factory?

astrofrog:
@brechmos-stsci - this isn't ideal as if cubeviz is installed, anyone using glue normally will also not have any other data viewers. I think resetting the data factories should only be done if cubeviz is started from the ``cubeviz`` command.

brechmos-stsci:
@astrofrog ok... so create a function in cubeviz data_factories to remove the defaults and then call that from the cubeviz.py startup function (?).

astrofrog:
Yes I think that makes sense!

brechmos-stsci:
Discussed with @astrofrog. I didn't realize the `setup()` function affects *any* installed glue app and not just in `cubeviz` where it is defined.  So, @astrofrog suggested moving the functionality from `setup()` to `main()` so as not to break other Glue apps.

cubeviz/cubeviz.py
```diff
@@ -41,9 +40,9 @@ def main(argv=sys.argv):
     args = parser.parse_known_args(argv[1:])
 
     # Store the args for each ' --data-configs' found on the commandline
-    global global_data_configuration
-    global_data_configuration['data_configs'] = args[0].data_configs if args[0].data_configs else []
-    global_data_configuration['data_configs_show'] = args[0].data_configs_show
+    data_configuration = {}
+    data_configuration['data_configs'] = args[0].data_configs if args[0].data_configs else []
+    data_configuration['data_configs_show'] = args[0].data_configs_show
```
astrofrog: Do you actually need the dictionary here, or can you just store in variables called ``data_configs`` and ``data_configs_show``?

Issue #171, opened 2018-01-18
Provide an easy way in the UI to create new regions and delete existing ones


astrofrog:
At the moment creating new regions and deleting them has to be done via the full glue UI in the left panel, so we need to think of a way to expose that to users. I'm a little reluctant to replicate the whole glue UI inside the cubeviz layout, so this will require careful thought.

nmearl:
Would it be possible to extend the selection modes available to include adding/removing selection regions? Something like:

![selection_modes_example](https://user-images.githubusercontent.com/4141126/35121289-e6c5120c-fc68-11e7-9a78-53b3f004ea7b.png)

astrofrog:
Yes, though we need to think how one picks which region to remove

drdavella:
This is currently being addressed in https://github.com/glue-viz/glue/pull/1515. Personally, for deleting regions, I like the idea of right clicking on a region and getting a context menu that provides the option to delete.

kassin:
Users like me like to hit the "delete" key. :)

drdavella:
New region creation is now the default mode when making selections in CubeViz. It is also possible to delete existing regions by right-clicking on them and selecting `Delete` from the dropdown menu.

PR #212, opened 2018-02-13
Provide custom viewer toolbar modes that create new ROIs

### LINKED ISSUE #134: All regions selected in images should always be displayed
If a user selects a region in the flux/err/DQ maps and then selects a second region, the first region should continue to display.  

Both spectra should display in specviz as is currently the case.

This should work for all regions selected, no matter how many.

drdavella:
This PR depends on updates to `glue` in this PR: https://github.com/glue-viz/glue/pull/1515. It should not be merged until the `glue` PR is merged.

The updates to `glue` allow ROI toolbar modes to indicate whether each click of the mode button should create a new ROI and subset. This is disabled by default, but this is the default behavior we want in `cubeviz`, so we create custom toolbar mode classes that override the `glue` default.

This partially addresses #171, and it fixes #134.

### DIFF
--- a/cubeviz/__init__.py
+++ b/cubeviz/__init__.py
@@ -9,3 +9,7 @@ This is an Astropy affiliated package.
 # ----------------------------------------------------------------------------
 from ._astropy_init import *
 # ----------------------------------------------------------------------------
+
+# This ensures that all custom toolbar modes are imported so that they are
+# registered by @viewer_tool and are available for use by CubeViz.
+from .toolbar_modes import *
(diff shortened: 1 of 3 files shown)


astrofrog:
@drdavella - I wonder whether this would be better implemented by adding a new subset mode for ``EditSubsetMode`` (in Glue) which would basically always create a new subset. We could then have a way to set what is the default and override that in Cubeviz?

drdavella:
@astrofrog that's an interesting idea. If you don't mind, I think I'd like to merge this as-is in conjunction with the `glue` PR, and then revisit this issue along with the other potential improvements I mentioned to you earlier.

drdavella:
Closing this since the ROI subset creation is now being handled entirely from within `glue` (https://github.com/glue-viz/glue/pull/1525, thanks @astrofrog!). It's possible that we'll require some other kind of minor change to `cubeviz` to flip a default setting in `glue`, but that remains to be seen.

PR #213, opened 2018-02-14
Use development version of spectral cube in environment file

drdavella:
Make sure that testers who use `test-environment.yml` get the latest development version of `spectral-cube`.

### DIFF
--- a/test-environment.yml
+++ b/test-environment.yml
@@ -30,5 +30,6 @@ dependencies:
   - glue-core>=0.12
   - pip:
     - asteval
+    - git+https://github.com/radio-astro-tools/spectral-cube#egg=spectral-cube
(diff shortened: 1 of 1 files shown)


test-environment.yml
```diff
@@ -30,5 +30,6 @@ dependencies:
   - glue-core>=0.12
   - pip:
     - asteval
+    - git+https://github.com/radio-astro-tools/spectral-cube#spectral-cube
```
robelgeda: Shouldn't it be `git+https://github.com/radio-astro-tools/spectral-cube#egg=spectral-cube`? (the `egg` part)
If not, it looks good to me.

drdavella: Yes, you're right! It was just a typo but it looks like it worked anyway. Will fix.

PR #226, opened 2018-02-16
Define Cubeviz data exporter

astrofrog:
I'm still working on trying to figure out how we make sure in a general sense that exported files in glue round-trip compared to the input files, but in the mean time I thought the easiest for CubeViz would be to define a custom data exporter that is cubeviz-specific.

This basically keeps a reference to the input HDUList and uses it when exporting the data. Components that no longer exist are removed from the HDUList, and new components are added. If the Primary HDU has a header but no data, that is preserved.

I can add tests once we add tests for the reading in.

### DIFF
--- a/cubeviz/data_factories/__init__.py
+++ b/cubeviz/data_factories/__init__.py
@@ -50,6 +50,14 @@ class DataConfiguration:
 
             self._data = cfg.get('data', None)
 
+    @property
+    def name(self):
+        return self._name
+
+    @property
+    def type(self):
+        return self._type
+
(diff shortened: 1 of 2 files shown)


brechmos-stsci:
I tested this by:

1) Load manga data and save it, then reload.  Worked

2) Load manga data, smooth and save all four components (flux, error, dq, smoothed flux), then reload. Worked

3) Load manga data, smooth and save only flux components (flux and smoothed flux), then reload.  Worked as well.
 
Going to merge when CI testing is done.

PR #246, opened 2018-02-20
Update astropy helpers submodule

drdavella:
~I expect this to fail for the time being.~


nmearl:
You ought to be able to just `cd` into the submodule and do a `git checkout d23a53f` (`d23a53f` being the latest pinned commit version), and then commit and push. Is it giving you issues?

drdavella:
@nmearl that's exactly what I did to create this PR.

drdavella:
However it was causing weird failures in another environment, so I was seeing if they would be picked up on CI. I actually think I might understand the issue now, and I think everything will be okay.

nmearl:
> and I think everything will be okay.

Ahh, the ideal. 😁

drdavella:
@nearl since you've already taken a quick look, do you mind approving this PR?

Issue #188, opened 2018-01-25
Collapse Cube Over Selected Wavelength Range


kassin:
Use case: collapsing a cube over a selected wavelength range.

@smei100 @bethanjames @drlaw1558 Please read and edit/comment as necessary as this is off of the top of my head and I'm likely not covering all use cases.

1. Select wavelength range over which to collapse in SpecViz, using the mouse.
   (For example, I select 3023 Angstroms - 3034 Angstroms with the mouse.)

2. Be able to edit the range by typing numbers in a dialog box or similar.
   (For example, I edit 3023 to be 3020 and 3034 to be 3035 using my keyboard.
   This is important because it's difficult to fine-tune selections with a mouse.)

3. Collapse the cube over this wavelength range.  The user should have the options to
  perform the collapse as a median, mean, or mode and with or without sigma clipping. 
  The user should have the option to collapse over the flux, error, and/or mask extensions.
  (For example, I wish to collapse flux, error, and mask over 3020 - 3035 Angstroms 
  using a median  and removing spaxels which deviate by more than 3-sigma.)

4. View and make measurements from the collapsed cube.  
  (For example, I wish to view the collapsed flux, error, and mask cubes
  in the three image panels across the top.)

(5. Save the collapsed cube.)  This is for future issue on saving files.

@nmearl @brechmos-stsci @drdavella

brechmos-stsci:
One step in this....

I talked with Nick and the idea is that cubeviz should do the collapsing (using spectral-cube for the WCS parts). So, for the *first* part of this I created a drop down to do cube collapsing:
![image](https://user-images.githubusercontent.com/18348847/36440061-c1f0e77e-163c-11e8-82cb-91fe3acb90bf.png)



Then if I choose the indexes in order to surround a specific peak I get:
![image](https://user-images.githubusercontent.com/18348847/36440088-d7e1e9ca-163c-11e8-9740-f3ade6ce66d7.png)

NEXT is to be able to take the ROI selected in specviz and send it back to cubeviz for the collapsing.

I think I will use the same modal so they can select which data component they want to process and the type of operation.

brechmos-stsci:
Let me finish the specviz part and then I'll show you.  Should be tomorrow 🤞

brechmos-stsci:
Question about the sigma clipping... I assume the user should be able to supply a number, N, such that the operation (e.g., sum, median, mean etc) will only be applied to spaxels that are within `mean(image) +/- N*sigma`?  

[...]

PR #247, opened 2018-02-21
Collapse cube

brechmos-stsci:
This PR implements the ability to collapse a cube along the spectral direction.  The primary modal is a new dialog under the "Data Processing" menu:
image

There are three ways to define a spectral range which we can collapse over:  1) input as indices of the data cube, 2) actual wavelength values or 3) define an ROI in the specviz viewer, then select the Collapse Cube modal and select the ROI in the "Region" Combo box.

image

Currently, this creates an overlay:
image

There is a WIP PR by @astrofrog that should allow this type of thing to be used in calculations but will have to wait for that PR to be complete.

This should complete #188


brechmos-stsci:
@nmearl and @drdavella, in particular, could you please try this out.   (Dan, you will have to update your specviz repo to be even with master).

brechmos-stsci:
This does not address the sigma clipping part in #188 so will have to look at that (maybe separately).

cubeviz/tools/collapse_cube.py
```diff
@@ -0,0 +1,359 @@
...
+
+        # Create start label and input box
+        self.start_label = QLabel("Start:")
+        self.start_label.setFixedWidth(100)
+        self.start_label.setAlignment((Qt.AlignRight | Qt.AlignTop))
+        self.start_label.setFont(boldFont)
+
+        self.start_text = QLineEdit()
+        self.start_text.setMinimumWidth(200)
+        self.start_text.setAlignment((Qt.AlignLeft | Qt.AlignTop))
```
drdavella: It would be very useful to populate both this and `end_text` with reasonable defaults. Maybe by default it should span the entire range of wavelengths? Or the middle 90%? I'll leave it up to someone else to decide what 'reasonable' means.

cubeviz/tools/collapse_cube.py
```diff
@@ -0,0 +1,359 @@
...
+    def calculate_callback(self):
+        """
+        Callback for when they hit calculate
+        :return:
+        """
+
+        # Grab the values of interest
+        data_name = self.data_combobox.currentText()
+        start_value = float(self.start_text.text().strip())
+        end_value = float(self.end_text.text().strip())
```
drdavella: There needs to be some error checking on the input strings here. Right now if these fields are empty, or if the values are invalid, we just get a `ValueError: could not convert string to float`. Maybe it would be useful to have the same kind of behavior that we implemented for the slice and wavelength value fields where invalid values turn red and are not accepted (actually it looks like this is done below, but the error handling logic needs to happen earlier).

cubeviz/tools/collapse_cube.py
```diff
@@ -0,0 +1,359 @@
...
+        if not 'Indices' in self.region_combobox.currentText():
+            wavelengths = np.array(self.parent._wavelengths)
+            start_value = np.argsort(np.abs(wavelengths - start_value))[0]
+            end_value = np.argsort(np.abs(wavelengths - end_value))[0]
+
+        self.error_label_text.setText(' ')
+        self.error_label_text.setStyleSheet("color: rgba(255, 0, 0, 128)")
+
+        # Sanity checks first
+        if not start_value and not end_value:
```
drdavella: This doesn't work. See comment above.

cubeviz/tools/collapse_cube.py
```diff
@@ -0,0 +1,359 @@
...
+            return
+
+        if start_value and end_value and start_value > end_value:
+            self.start_label.setStyleSheet("color: rgba(255, 0, 0, 128)")
+            self.end_label.setStyleSheet("color: rgba(255, 0, 0, 128)")
+            self.error_label_text.setText('Start value must be less than end value')
+            return
+
+        # Set the start and end values if they are not set.
+        if not start_value:
```
drdavella: This doesn't work. See comment about error handling above.

cubeviz/tools/collapse_cube.py
```diff
@@ -0,0 +1,359 @@
...
+        self.error_label_text.setStyleSheet("color: rgba(255, 0, 0, 128)")
+
+        # Sanity checks first
+        if not start_value and not end_value:
+            self.start_label.setStyleSheet("color: rgba(255, 0, 0, 128)")
+            self.end_label.setStyleSheet("color: rgba(255, 0, 0, 128)")
+            self.error_label_text.setText('Must set at least one of start or end value')
+            return
+
+        if start_value and end_value and start_value > end_value:
```
drdavella: Just tried testing this by putting `1` in the start value box, and `0` in the end value box, and did not see any kind of error message.

robelgeda:
Dont forget the escape key handler. I think this snippet from moments.py will work here too:
```
def keyPressEvent(self, e):
        if e.key() == Qt.Key_Escape:
            self.cancel_callback()
```

drdavella:
@robelgeda, @brechmos-stsci it seems like at some point we might want to implement a parent class for all of these dialogs that provides `cancel_callback` and `keyPressEvent` so that we don't have to keep redefining them for each subclass.

robelgeda:
@drdavella I really like that idea. I will make an issue

cubeviz/tools/collapse_cube.py
brechmos-stsci: will be fixed in the next commit

cubeviz/tools/collapse_cube.py
brechmos-stsci: will be fixed in the next commit

cubeviz/tools/collapse_cube.py
brechmos-stsci: will be fixed in the next commit

cubeviz/tools/collapse_cube.py
brechmos-stsci: will be fixed in the next commit

cubeviz/tools/collapse_cube.py
brechmos-stsci: I am going to leave this for now and let them put it in. I think this will force the idea that using the specviz ROI might be the best method (??)

PR #263, opened 2018-03-03
Update index.rst

kassin:


brechmos-stsci:
#267  will be used instead... closing this

PR #273, opened 2018-03-08
More shortcuts

javerbukh:
# Added the following shortcuts

- A: move slice 1 to the left
- D: move slice 1 to the right
- F: freeze coordinates while hovering over a plot
- P: pull coordinates (frozen or not) into users clipboard, where they can be pasted anywhere

### DIFF
--- a/cubeviz/image_viewer.py
+++ b/cubeviz/image_viewer.py
@@ -187,7 +187,8 @@ class CubevizImageViewer(ImageViewer):
         # Connect matplotlib events to event handlers
         self.statusBar().messageChanged.connect(self.message_changed_callback)
         self.figure.canvas.mpl_connect('motion_notify_event', self.mouse_move)
-        self.figure.canvas.mpl_connect('axes_leave_event', self.mouse_exited)
+        #self.figure.canvas.mpl_connect('axes_leave_event', self.mouse_exited)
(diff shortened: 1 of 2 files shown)


hcferguson:
Thinking aloud I might suggest these alternatives:

  - Tab, space or right arrow: Move one slice to the right
  - shift-tab, backspace, or left arrow: Move one slice to the left
  - C: Pull coordinates into the clipboard

These seem more intuitive to me based on using ds9 and imexam in the past.

A couple questions/comments:
  - Are left and right consistent in moving toward longer wavelength (right?) and shorter (right?) even if the units are in frequency? I don't have any strong opinion on whether this is desired, just curious.

  - @robelgeda If we're not doing so already, it's worth thinking about the formatting of the coordinates when we save them so that people can easily use them, e.g. in Astropy. For example by saying 
 ``` c = SkyCoord(pasted output)```

brechmos-stsci:
@hcferguson There are currently some issues or difficulties with trying to grab non-alphabetic (e.g., arrow keys) right now from cubeviz.  It was going to be a lot more work to get arrow keys (deeper in glue).  So I had suggested that we get the basic functionality in there for now (as the ticket has been open for a while) and then come back to a larger look at the actual key bindings using arrow keys etc.  And, in particular, as there is other work we need to get done too. 

It is a great point about wavelength vs frequency.  Currently, I believe the left is subtract one from the index and right is add one to the index.

PR #286, opened 2018-03-12
removed unused menu items, will add in when the functionality is added

brechmos-stsci:
Removed unused menu items so as not to confuse people using it. We can add them back in when the functionality exists.


drdavella:
Is there an issue tracking these features?

PR #291, opened 2018-03-13
WIP: Updating the update script for the testers

### LINKED ISSUE #290: (New) update script for testers
The update script was based on the astroconda-dev package system that builds packages from each repo's master environment. 

There appears to be a continuum of types of users of cubeviz.  Developers who will likely use master of each repo (or something close to it). Testers who we want to use specific, reasonably new, versions of some repos and then end users who will use, essentially tagged versions of repos.

For the testers, we need to setup an environment that has specific versions of the "big 6" repositories: cubeviz, specviz, specutils, spectral-cube, glue and astropy.  Some of the repos are fast changing and we may not necessarily want the master version for the testers to do their testing.

So..... I am going to update the update script (it's a meta thing) that will pull down and run a script that has specific commit hashes in it in order to lock the testing environment to what *we* want. This way we can test it beforehand and we will know exactly what version of the "big 6" they are at.

cc @kassin @drdavella

brechmos-stsci:
All very meta...  The scripts here were created and updated in order to force specific versions of the big 5 packages on the testers for their environment.  This will use pip to install specific commit hashes of the primary packages.

This is to fix #290

### DIFF
--- a/scripts/update_cubeviz_test_env
+++ b/scripts/update_cubeviz_test_env
@@ -20,9 +20,5 @@ output_file=`mktemp`
 # https://stackoverlfow.com/questions/14411103/check-for-existence-of-wget-curl
 wget -O - --no-check-certificate $download_location > $output_file || curl -o $output_file $download_location
 
-echo "-------------"
-cat $output_file
-echo "-------------"
-
 echo "Running the update script $output_file"
 sh $output_file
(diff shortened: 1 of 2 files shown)


brechmos-stsci:
@drdavella When you have a few minutes, could you please test this on your machine?   You will have to edit the variable `download_location` in `scripts/update_cubeviz_test_env` to be `https://raw.githubusercontent.com/brechmos-stsci/cubeviz/tester-script/scripts/update_cubeviz_test_env_pip` before you run it.

brechmos-stsci:
Note, too, this will not work on Windoze, which I think is fine for now as I don't believe any of our testers are using it.

PR #293, opened 2018-03-13
Added test (gasp) for update script

brechmos-stsci:
Added test for the new update script to make sure the commit hashes exist.


brechmos-stsci:
Next, I want to move this type of dict thing to the update script which will simplify this test script. But, for now...

drdavella:
Merging despite unrelated test failures on Windows.

PR #297, opened 2018-03-13
Get rid of named viewers in split layout

drdavella:
This pays down some code debt and makes the split layout more general, allowing us to potentially support an arbitrary number of viewers in the future. We still retain the `single_viewer` named viewer since it is convenient to do so and does not cause any additional loss of generality.

### DIFF
--- a/cubeviz/layout.py
+++ b/cubeviz/layout.py
@@ -82,21 +82,19 @@ class CubeVizLayout(QtWidgets.QWidget):
         self.ui = load_ui('layout.ui', self,
                           directory=os.path.dirname(__file__))
 
-        self.all_views = []
+        self.cube_views = []
 
         # Create the cube viewers and register to the hub.
         for _ in range(DEFAULT_NUM_SPLIT_VIEWERS + 1):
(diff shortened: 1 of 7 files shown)


cubeviz/layout.py
```diff
@@ -74,23 +77,20 @@ def __init__(self, session=None, parent=None):
...
+        # Create specviz viewer and register to the hub.
+        self.specviz = WidgetWrapper(SpecVizViewer(self.session), tab_widget=self)
+        self.specviz._widget.register_to_hub(self.session.hub)
 
-        # TODO: determine whether to rename this or get rid of it
-        self.cube_views = self.all_views
+        self.single_view = self.cube_views[0]
         self.split_views = self.cube_views[1:]
 
         self._synced_checkboxes = [
```
drdavella: While this was not generalized, we plan to move the sync checkboxes to the toolbars of the viewers themselves, so it doesn't make sense to fix this here yet.

brechmos-stsci:
There is a conflict (now).  Could you fix it then I'll merge

PR #307, opened 2018-03-14
Changed Custom -> Other

brechmos-stsci:
This was requested by @kassin.

The word "Custom" suggested something that could be changed about the components.


brechmos-stsci:
@robelgeda If you could ck and approve please.

robelgeda:
This doesn't affect anything

PR #310, opened 2018-03-15
correction on the dir name

brechmos-stsci:


brechmos-stsci:
This was needed as the directory path was an old incorrect one.

PR #315, opened 2018-03-15
remove all remnants of the old version (will now be stored in the gists)

brechmos-stsci:
The update stuff will now be stored in the gists as it was too painful to have in here.


brechmos-stsci:
@drdavella , if the new way is to do the gist scripts, then I think we can delete all this, right?

drdavella:
Yes I think so. You'll need to remove the test as well.

PR #316, opened 2018-03-15
Display ROI/subset stats

drdavella:
This PR introduces a feature that displays basic statistics about selected ROIs in the viewer:

image

Currently the stats include mean and standard deviation of the flux values under the selected region for the current slice. If it is useful to do so, the stats could include mean and standard deviation of flux values through the entire cube under the selected region.

This also adds a top-level configuration option that determines whether the stats should be displayed or not:

image

Currently no units are displayed. This would probably be a useful addition, but unfortunately there is limited real estate.

Comments and suggestions are welcome.

### DIFF
--- a/cubeviz/image_viewer.py
+++ b/cubeviz/image_viewer.py
@@ -191,6 +191,8 @@ class CubevizImageViewer(ImageViewer):
         self._toggle_3d = False  # True if we just toggled from 2D to 3D
 
         self._stats_axes = None
+        self._stats_visible = True  # Global configuration setting
+        self._stats_hidden = False  # Internal configuration: are stats hidden?
         self._component = None # Keep track of name of currently displayed component
         self._subset = None # Keep track of currently active subset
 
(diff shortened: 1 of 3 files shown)


cubeviz/image_viewer.py
```diff
@@ -516,6 +591,9 @@ def update_slice_index(self, index):
         if self.is_contour_active:
             self.draw_contour()
 
+        if self._stats_axes is not None:
+            self.update_stats()
+
     def fast_draw_slice_at_index(self, index):
         """
```
robelgeda: The stats will not update in `fast_draw_slice_at_index`. Does this work well when you slide?

cubeviz/image_viewer.py
```diff
@@ -218,6 +227,72 @@ def get_data_layer_artist(self, layer=None, layer_state=None):
...
+        return data.mean(), data.std()
+
+    def draw_stats_axes(self, component, subset):
+
+        self._component = component
+        self._subset = subset
+
+        mu, sigma = self._calculate_stats(component, subset)
+
+        if self._stats_axes is None:
```
robelgeda: What if you have multiple ROIs?

drdavella: Good question: the stats are only displayed for the "active" ROI. If you have multiple ROIs, clicking on each one should update the stats to reflect that ROI. Currently the "active" ROI is only shown in the layer tree widget, but we hope to modify glue to show some kind of marker on the viewer itself indicating which ROI is active.

cubeviz/image_viewer.py
drdavella: The box doesn't disappear, which is cool! Personally I think it's fine to update the stats only when the slider is released. It might not be that much of a performance hit to do it while sliding, but it doesn't seem necessary.

robelgeda: agreed

astrofrog:
It would be nice to have a way to show stats that doesn't depend on showing it inside a Matplotlib figure - this won't work for the specviz viewer nor for the 3D viewers. Maybe in future we could instead have the stats show up in a kind of floating 'inspector-style' dialog? (and this could be implemented at the glue level instead)

PR #317, opened 2018-03-15
Do not automatically display overlays

### LINKED ISSUE #309: Overlay should not be shown by default, 2D component should popup initially
When the collapse cube or moment maps are calculated, the overlay should not be displayed, but the 2D component should be displayed by default (maybe in the active viewer / or first viewer).

drdavella:
With this PR, overlays resulting from moment maps and collapsed cubes are no longer displayed automatically. Instead, the last active cube viewer is automatically populated with the result of the computation.

This fixes #309.

### DIFF
--- a/cubeviz/layout.py
+++ b/cubeviz/layout.py
@@ -326,7 +326,7 @@ class CubeVizLayout(QtWidgets.QWidget):
         operation_handler.exec_()
 
     def add_new_data_component(self, component_id):
-        self.update_active_view(component_id)
+        self.display_component(component_id)
 
     def remove_data_component(self, component_id):
         pass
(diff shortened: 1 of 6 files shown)


cubeviz/layout.py
```diff
@@ -449,13 +441,23 @@ def change_viewer_component(self, view_index, component_id, force=False):
...
         if combo.currentIndex() == component_index and force:
             combo.currentIndexChanged.emit(component_index)
         else:
             combo.setCurrentIndex(component_index)
 
+    def display_component(self, component_id):
+        """
+        Displays data with given component ID in the active cube viewer.
+        """
+        view_index = self.cube_views.index(self._active_cube)
```
robelgeda: I need to ask the testers about this actually. (I think only the first viewer should update not the active one)

PR #323, opened 2018-03-19
Minor tweaks to interactions between spectral-cube and cubeviz

nmearl:
This PR includes the ability to perform calculations using the entire cube as input in addition to applying a function allowing particular axes in a `SpectralCube`.


brechmos-stsci:
@nmearl I'll test it this morning.  Could you please correct the failing tests? (I think just pep8)

PR #331, opened 2018-03-19
Redshift

### LINKED ISSUE #193: Converting from Observed to Rest-frame Wavelengths in SpecViz
From @bethanjames:

"It would be very useful for the user to apply a redshift value near the wavelength selection tool, that way you could offset the wavelength array and move through the cube in the rest-frame wavelengths."

@nmearl I know this is implemented in the stand-along SpecViz.  Do you have a suggestion as how it would work in CubeViz?

@brechmos-stsci

brechmos-stsci:
Redshift is basically done.  It was implmented in the units.py as that contains the main wavelength work.  

For the message to come from specviz back to cubeviz, we need https://github.com/spacetelescope/specviz/pull/377 or something similar.

This fixes #193


brechmos-stsci:
@drdavella Could you review this PR when you get a chance.  It should be good (short of any test errors).

cubeviz/controls/slice.py
```diff
@@ -31,6 +35,11 @@ def __init__(self, cubeviz_layout):
         self._slice_slider.setEnabled(False)
         self._slice_textbox.setEnabled(False)
         self._wavelength_textbox.setEnabled(False)
+        self._wavelength_textbox_label.setWordWrap(True)
+
+        # This should be used to distinguised between observed and rest wavelengths
+        # We are not going to enforce what the name should be at this level.
+        self._wavelength_label_text = 'Obs Wavelength'
```
drdavella: Since this string and `Rest Wavelength` are each used in a few different places, it might be worthwhile to define them in global constants somewhere.

cubeviz/controls/slice.py
```diff
@@ -272,6 +307,17 @@ def specviz_wavelength_slider_change(self, event=None, pos=None):
             deactivate_flag = True
             self._slider_flag = True
 
+        # The "pos" value coming from specviz appears to be related to the
+        # index in the observed wavelength and so if there is a redshift
+        # then we need to convert the pos to the rest wavelength position.
+        if self._cv_layout._units_controller and not self._cv_layout._units_controller.redshift_z == 0.0:
+            #rest_wavelength = self._cv_layout._units_controller._original_wavelengths[pos] / (1 + self._cv_layout._units_controller.redshift_z)
```
drdavella: Is there any reason to keep this line around?

cubeviz/controls/units_ui.py
```diff
@@ -0,0 +1,202 @@
...
+        self.currentKernel = None
+
+        self.createUI()
+
+    def createUI(self):
+        """
+        Create the popup box with the calculation input area and buttons.
+
+        :return:
+        """
```
drdavella: I've mentioned this before but I think it would be a good idea in the long term if we redefined all of these UI layouts in `.ui` files using Qt Designer.

cubeviz/controls/slice.py
brechmos-stsci: Added globals.

cubeviz/controls/slice.py
brechmos-stsci: Removed

cubeviz/controls/units_ui.py
brechmos-stsci: agreed. we can do that part of a cleanup.

PR #332, opened 2018-03-19
better description if file does not exist

brechmos-stsci:
This will tell the user that the file does not exist.  Without this it says that cubeviz could not read the file which is a little too nebulous.

### DIFF
--- a/cubeviz/cubeviz.py
+++ b/cubeviz/cubeviz.py
@@ -1,6 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import sys
 import argparse
+import os
 
 from glue.app.qt import GlueApplication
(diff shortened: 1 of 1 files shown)


cubeviz/cubeviz.py
```diff
@@ -61,6 +62,12 @@ def main(argv=sys.argv):
 
     datafiles = args[0].data_files
 
+    # Check to make sure each file exists and raise an Exception 
+    # that will show in the popup if it does not exist. 
+    for x in datafiles:
+        if not os.path.isfile(x):
+            raise IOError('The file {} does not exist'.format(x))
```
drdavella: Not a big deal, but it might be useful to put the filename in quotes here.

Issue #304, opened 2018-03-14
Support a more flexible layout for cube viewers


drdavella:
It would be nice to untie ourselves from a fixed three-viewer layout and to allow users flexibility in creating/removing new viewers as necessary and positioning viewers wherever they want.

See https://github.com/glue-viz/glue/issues/837 for upstream changes that could help us with this.

kassin:
"free the viewers" ;)

astrofrog:
I like the way Jupyter Lab deals with customizing the layout:

![flexible](https://user-images.githubusercontent.com/314716/37453117-01fb047e-2837-11e8-859c-e20ab6c39e89.gif)

It'd be nice if we could come up with a way to do it in Qt.

drdavella:
As @hcferguson mentioned in #116, we should also provide an option allowing a viewer to fill the current canvas.

PR #343, opened 2018-03-28
Refactor to use messaging for index and wavelength updates

drdavella:
This is a fairly major refactor of cubeviz infrastructure (done in collaboration with @brechmos-stsci). We have introduced messages for slice changes, and for wavelength, wavelength unit, and redshift changes. This has simplified the implementation of  several classes and has removed cross dependencies that made the code harder to reason about.

This will facilitate future changes that will be required for making the layout more flexible (as described in issue #304).

This PR is based on #337 and will need to be rebased after that is merged.


drdavella:
Maybe some of this commit history should be squashed up before merging?

PR #344, opened 2018-03-30
WIP: Loading of data using ifucube in order to do correction

brechmos-stsci:
**Do not merge (yet), this code should be fixed to work for most IFU cubes we have**

There are many IFU cubes that aren't quite correct and will not load due to requirements in packages used by cubeviz.  (e.g,. if CTYPE3 is not correct then spectral-cube will not be able to determine the spectral axis). 

The `ifucube.py` file is used to load the data and do sanity checks and corrections on the IFU cube so cubeviz can read all the data files. 

[After our first release the ifcube.py should be pulled out of here into it's own package and then cubeviz should just import from the new package. This way we can do fixes and releases of a new IFUCube package without having to re-release cubeviz.  This would also make (automated) testing of data cubes a lot easier and would make it (probably) cleaner to implement a GUI wizard in the far future.]

A few things that need to be fixed in this:
   - [ ] `CTYPE3` is not being corrected, umm, correctly in files such as `/astro/3/jwst_da_sprint_testdata/IFU_datacubes/MUSE/DATACUBE_HDFS_v1p0.fits.gz`. We maybe should include a `w = WCS(header)` to check that it works correctly.
  - [ ] Will need to check if CTYPE{1,2,3} order is dependent on the order of the data axes
  - [ ] Remove assumption that CTYPE3 is the wave axis (unless spectral cube assumes that, then correct)
  - [ ] Unit correction... one file has 'Angstroms' but it should be 'Angstrom' and so is failing.. maybe some simple unit corrections (??)


brechmos-stsci:
Closed in favor of #374

Issue #328, opened 2018-03-19
Arithmetic calculates in cubeviz


brechmos-stsci:
@astrofrog If the goal is to move to use the Glue "Derived Component Editor":
![image](https://user-images.githubusercontent.com/18348847/37596766-999f25c0-2b53-11e8-9a7b-d0efdfb3c869.png)

Could you change the menu "Arithmetic Operations" to go right to the "Derived Component Editor" (not the Add/Edit Components)
![image](https://user-images.githubusercontent.com/18348847/37596789-aa8c1488-2b53-11e8-8039-f25b140d5e62.png)

astrofrog:
@brechmos-stsci - there is an issue open for this already: https://github.com/spacetelescope/cubeviz/issues/284 - could you re-post this comment there? I'll try and work on this today or tomorrow.

astrofrog:
I'm working on this today

astrofrog:
Preliminary work required for this: https://github.com/glue-viz/glue/pull/1637

Issue #284, opened 2018-03-09
Combine Glue equation editor and Cubeviz arithmetic editor


astrofrog:
### Background

Glue has a data component manager accessible via 'Add/Edit components':

<img width="517" alt="screen shot 2018-03-09 at 22 02 52" src="https://user-images.githubusercontent.com/314716/37232152-a725d920-23e5-11e8-86f3-c5dff3c9d0f3.png">

Clicking on 'New derived component' at the bottom brings up an equation editor:

<img width="420" alt="screen shot 2018-03-09 at 22 02 56" src="https://user-images.githubusercontent.com/314716/37232160-b1b871ea-23e5-11e8-814e-3b9c380840a8.png">

At the same time, cubeviz has its own arithmetic editor:

<img width="654" alt="screen shot 2018-03-09 at 22 03 40" src="https://user-images.githubusercontent.com/314716/37232171-c710508a-23e5-11e8-90fc-98ea77b2a978.png">

There are pros/cons to each:

**Glue pros:**

- Component manager allows users to go back and edit existing components they've created
- Component manager allows users to remove components they don't need anymore
- The equation editor has an easy way to insert components
- The equation editor has syntax highlighting

**Cubeviz editor pros:**

- Easier to find through the drop-down menu (it's not obvious that users need to click on add/edit components in the toolbar)
- The examples are nice

### Proposal

I propose that we try and transition to somehow using the glue equation editor, but we can subclass it to add the examples (and make them copy/pastable). I'd also like to think about whether we can expose the component manager so that users can go back and edit previously added components. However, I acknowledge that the UI of the built-in glue component manager and how to find it is not obvious. So I'd be interested in what people think - how could we improve the built-in glue one (besides adding examples) to the point where we could use this instead of the current cubeviz equation editor?

PR #347, opened 2018-04-03
Use built-in glue arithmetic editor

astrofrog:
This replaces the arithmetic editor in cubeviz with the one in glue, which has recently been cleaned up. Users choosing 'Arithmetic operations' in the drop-down (or via the toolbar icon from glue) will now be presented with a list of current arithmetic components:

screen shot 2018-04-03 at 16 46 39

and a way to define/edit them:

screen shot 2018-04-03 at 16 46 36

This provides the following improvements to cubeviz:

* Users can go back and edit existing definitions or remove definitions
* Syntax highlighting helps users identify issues with the expression
* The combo box makes it easier to insert components

I've ported some of the advantages of the original cubeviz editor over to glue, including the examples.

@brechmos-stsci - note that the dialog above defaults to FLUX for the examples.

**Note:** this requires the very latest version of glue to work.

Fixes https://github.com/spacetelescope/cubeviz/issues/328
Fixes https://github.com/spacetelescope/cubeviz/issues/284


brechmos-stsci:
@astrofrog Hmmm. I did a pull on Glue master and went into the branch but am seeing an error when I startup cubeviz:
image

Looks different than the test error but maybe related?

astrofrog:
The entry points have changed in glue so running ``python setup.py develop`` or ``python setup.py install`` is required.

PR #359, opened 2018-04-04
Default YAML file

javerbukh:
I created a yaml file called default that only calls the leaf function "has_data." has_data then returns whether or not the file has data. The priority is currently set for 1000.

### DIFF
--- a/cubeviz/data_factories/__init__.py
+++ b/cubeviz/data_factories/__init__.py
@@ -139,8 +139,11 @@ class DataConfiguration:
                             c = data.get_component(component_name)
                             c.units = self.get_units(hdu.header)
                     else:
-                        component_name = os.path.basename(data_filename)
+                        # Creates a unique component name
+                        component_name = str(ii)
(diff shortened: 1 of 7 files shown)


cubeviz/data_factories/__init__.py
```diff
@@ -229,6 +231,14 @@ def _any(self, conditionals):
     # Leaf processing
     #
 
+    def _has_data(self):
+        """
+        Used for miscellaneous cubes that we have not planned for by just making sure data in the cube exists. This
+        is currently only in use in the default.yaml file
+        :return:
+        """
+        return hasattr(self._fits[0], 'data')
```
brechmos-stsci: @javerbukh I think it might be safer to see if any HDU has an attribute data rather than just the first.  I think, too, that it might be better to see if it exists (as you have) AND that the len(self._fits[x].data) == 3. I think it is fair to say we are expecting a 3D cube of data.

PR #360, opened 2018-04-04
fixed the spacing on the UI

brechmos-stsci:
Fixed the spacing on the redshift / wavelength UI.  Not perfect but better.


brechmos-stsci:
@drdavella could you do a quick test of this please.

PR #362, opened 2018-04-04
Add median to ROI stats

### LINKED ISSUE #321: Add median to displayed statistics for selected ROI

drdavella:
This resolves #321. It also fixes a bug that was introduced by #343 that causes the wrong ROI stats to be displayed when the slider is released.


drdavella:
image

Issue #235, opened 2018-02-19
Specviz smoothed data should be annotated with more information


brechmos-stsci:
The data coming back from specviz (related to https://github.com/spacetelescope/cubeviz/pull/214) is called "FLUX [Spectrally Smoothed]" and has an increment if the component already exists.

It would be better to include information of how the component was spectrally smoothed (e.g., "FLUX [spectrally smoothed, Gaussian, sigma = 7 pixels]".

PR #365, opened 2018-04-05
changed description to include parameters of smoothing

brechmos-stsci:
I changed the spectral smoothing description to include parameters of the smoothing.  So it now looks like:
image

(previously it just said "Flux Smooth".

This fixes https://github.com/spacetelescope/cubeviz/issues/235.


brechmos-stsci:
This fixes https://github.com/spacetelescope/cubeviz/issues/235.

Issue #282, opened 2018-03-09
Collapse cube wavelength range bug


charlottenosam:
**v0.1.dev534**

When trying to use the Data Processing / Collapse Cube to collapse a cube along wavelength axis between 2 wavelength values I get the following error message:
![collapsecube](https://user-images.githubusercontent.com/5003032/37220888-df1b8b90-237c-11e8-8cba-9b0dd68f3c9d.png)

If I switch to Region: Indices, or if I select a region in SpecViz, and then use Collapse Cube, there are no problems.

brechmos-stsci:
Need to display units in the Start and End Wavelength labels.

brechmos-stsci:
The GUI has been completely redone so closing this.

PR #368, opened 2018-04-06
Collapse cube UI refactor

### LINKED ISSUE #314: Collapse cube sigma clipping interface
@smei100 brought up a question about the sigma clipping for the collapse cube. The upper and lower clipping values were not clear.  She would prefer to have just a single sigma value to clip from.  

@charlottenosam also brought up the idea that clipping based on SNR might be helpful too.

As a slight aside, we might want to check the wavelength values and give a hint as to why a value did not work and/or suggested values.

brechmos-stsci:
This is a major refactor of the collapse cube UI.  One can now choose no sigma clipping, simple sigma clipping (one sigma value) or advanced sigma clipping (that provides an interface to `astropy.stats.sigma_clip`).

If one provides a region defined in specviz then one can switch to "Custom (Wavelengths)" and the wavelength values can then be tweaked.  Also, one can switch to "Custom (Indices)" and tweak the indices instead.

For the simple sigma clipping this just calls `astropy.stats.sigma_clip(data, sigma=sigma_provided, axis=0)`.  

For the advanced sigma clipping all provided values are passed to `astropy.stats.sigma_clip`

Fixes #314  #282


cubeviz/tools/collapse_cube.py
```diff
@@ -311,187 +228,321 @@ def hide_start_end(self, dohide):
...
-        ts = start_index if 'Indices' in self.region_combobox.currentText() else self.wavelengths[start_index]
-        self.start_text.setText('{}'.format(ts))
-
-        te = end_index if 'Indices' in self.region_combobox.currentText() else self.wavelengths[end_index]
-        self.end_text.setText('{}'.format(te))
+#        ts = start_index if 'Indices' in self.region_combobox.currentText() else self.wavelengths[start_index]
+#        self.ui.start_input.setText('{}'.format(ts))
+#
+#        te = end_index if 'Indices' in self.region_combobox.currentText() else self.wavelengths[end_index]
+#        self.ui.end_input.setText('{}'.format(te))
```
drdavella: Is there any reason to keep this code around?

cubeviz/tools/collapse_cube.py
```diff
@@ -311,187 +228,321 @@ def hide_start_end(self, dohide):
...
-
         # Get the start and end wavelengths from the newly created spectral cube and use for labeling the cube.
         # Convert to the current units.
-        start_wavelength = new_wavelengths[0].to(self.wavelength_units)
-        end_wavelength = new_wavelengths[-1].to(self.wavelength_units)
+        start_wavelength = self.wavelengths[start_index]
+        end_wavelength = self.wavelengths[end_index]
 
-        label = '{}-collapse-{} ({:0.3}, {:0.3})'.format(data_name, operation,
+        label = '{}-collapse-{} ({:0.4}, {:0.4})'.format(data_name, operation,
```
drdavella: There is currently not enough precision in the wavelength values to prevent duplicated data components.

cubeviz/tools/collapse_cube.py
```diff
@@ -311,187 +228,321 @@ def hide_start_end(self, dohide):
...
+            self.ui.start_label.setStyleSheet("color: rgba(255, 0, 0, 128)")
+            self.ui.error_label.setText('Start wavelength is out of range.')
+            self.ui.error_label.setVisible(True)
+
+            start_wavelength = None
+            
+
+        if end_wavelength > self.wavelengths[-1]:
+            self.ui.end_label.setStyleSheet("color: rgba(0, 0, 255, 128)")
+            self.ui.error_label.setText('End wavelength is out of range, setting to end point.')
```
drdavella: I set `Start Wavelength` to 1000 and `End Wavelength` to 10, and this is the error message I get, which is not what I would expect. I think the `start_wavelength < end_wavelength` test should take priority over these?

cubeviz/tools/collapse_cube.py
```diff
@@ -311,187 +228,321 @@ def hide_start_end(self, dohide):
...
+
+            start_wavelength = self.wavelengths[0]
+            
+        if start_wavelength > self.wavelengths[-1]:
+            self.ui.start_label.setStyleSheet("color: rgba(255, 0, 0, 128)")
+            self.ui.error_label.setText('Start wavelength is out of range.')
+            self.ui.error_label.setVisible(True)
+
+            start_wavelength = None
+
```
drdavella: There's currently no check for negative wavelength or index values. If I put in something like `-100` for either `Start Wavelength` or `Start Index`, the cube operation seems to "work".

drdavella: If I do `Start Wavelength/Index` as `-100` and `End Wavelength/Index` as `-1`, I get `End wavelength/index out of range`.

cubeviz/tools/collapse_cube.py
```diff
@@ -311,187 +228,321 @@ def hide_start_end(self, dohide):
...
+        log.debug('    simple_sigma {}'.format(simple_sigma))
+
+        try:
+            simple_sigma = float(simple_sigma)
+        except:
+            self.ui.simple_sigma_label.setStyleSheet("color: rgba(255, 0, 0, 128)")
+            self.ui.error_label.setText('Sigma must be a floating point number.')
+            self.ui.error_label.setVisible(True)
+            return None
+
```
drdavella: Is it okay to allow `sigma <= 0` here?

cubeviz/tools/collapse_cube.py
```diff
@@ -311,187 +228,321 @@ def hide_start_end(self, dohide):
...
+        sigma_iters = self.ui.advanced_sigma_iters_input.text().strip()
+
+        if len(sigma) > 0:
+            try:
+                sigma = float(sigma)
+            except:
+                self.ui.advanced_sigma_label.setStyleSheet("color: rgba(255, 0, 0, 128)")
+                self.ui.error_label.setText('Sigma must be a floating point number.')
+                self.ui.error_label.setVisible(True)
+                return None
```
drdavella: I'm not that familiar with use cases here, but why do I need `Sigma` in addition to `Sigma Lower` and `Sigma Upper`? I get an error unless I provide `Sigma` in addition to the upper and lower, but it's not at all clear why that's necessary.

cubeviz/tools/collapse_cube.py
```diff
@@ -311,187 +228,321 @@ def hide_start_end(self, dohide):
...
+            try:
+                sigma_upper = float(sigma_upper)
+            except:
+                self.ui.advanced_sigma_upper_label.setStyleSheet("color: rgba(255, 0, 0, 128)")
+                self.ui.error_label.setText('Upper sigma must be a floating point number.')
+                self.ui.error_label.setVisible(True)
+                return None*4
+        else:
+            sigma_upper = None
+
```
drdavella: There's no check to make sure `sigma_lower < sigma_upper`, or to make sure that either of these values are `<= 0`.

cubeviz/tools/collapse_cube.py
hcferguson: This is just for convenience for the default case.  ```astropy.stats.sigma_clip``` does it this way, over-riding Sigma if separate ```Sigma_Lower``` or ```Sigma_Upper``` are provided (not sure what happens if only one is provided).

cubeviz/tools/collapse_cube.py
hcferguson: I don't think a check for <= 0 is worth doing here. Someone could easily subtract sky and end up wanting ```sigma_lower <= 0.``` It would be unusual to want ```sigma_upper <= 0.``` 

It's good to check that sigma_lower < sigma_upper...but I wonder if ```astropy.stats.sigma_clip``` does that, in which case trapping the error might be a better approach than doing the check before passing the parameters.

cubeviz/tools/collapse_cube.py
brechmos-stsci: The request was to have the abillity to put a value in too high or too low and that it peg to the min or max value. If the start value is too high or the end value is too low *then* there is an error as noted.

cubeviz/tools/collapse_cube.py
brechmos-stsci: I switched the order to see the start> end message.

cubeviz/tools/collapse_cube.py
brechmos-stsci: I added a check here.

brechmos-stsci:
@drdavella I went through and believe I addressed your queries.  Please re-check.

cubeviz/tools/collapse_cube.py
drdavella: I'm not sure this fixes the issue in a really robust way...

cubeviz/tools/collapse_cube.py
drdavella: Shouldn't there still be a check for negative values, though?

cubeviz/tools/collapse_cube.py
drdavella: For the purposes of this UI, shouldn't I be allowed to omit sigma entirely if I provide both upper and lower bounds?

drdavella: I think that if `sigma` has a default, then the default should automatically populated into our UI. I think the same should be true of Start Wavelength and End Wavelength, and Start Slice and End Slice, otherwise it's not immediately obvious what's going to happen when I don't enter any values.

cubeviz/tools/collapse_cube.py
drdavella: Now if I put in a value for sigma, but not for Start Wavelength or End Wavelength, then the "End Wavelength" label turns purple but nothing else happens. Shouldn't I be able to apply sigma clipping to the entire cube by default without explicitly specifying start and end points?

brechmos-stsci: added a message in the final dialog to show the start and end wavelengths were used. removed the purple thing.

PR #370, opened 2018-04-06
Pass cubeviz layout reference to specviz 

nmearl:
Allows specviz to add new data cube components and overlays.


nmearl:
Go ahead and try again.

PR #375, opened 2018-04-07
Improve display of slice and ROI statistics

drdavella:
This PR provides a (hopefully) improved way of displaying slice and ROI statistics:

image

The statistics can be hidden using a view menu option:

image

The following should be improved/implemented before merging:
- [ ] Display the name of the active subset next to the ROI stats
- [ ] Emphasize the stats display somewhat so that it doesn't get lost in the UI


drdavella:
image

Issue #326, opened 2018-03-19
Plot Layers Color/Opacity Sync make not default


brechmos-stsci:
The "Sync" button is default to sync, but should be default not to sync. That way when the user changes the color/opacity it won't do it for all 3 windows (not ideal)

![image](https://user-images.githubusercontent.com/18348847/37595932-e704ad9c-2b50-11e8-8368-01b37b512582.png)

@kassin

brechmos-stsci:
There was another comment that maybe the "Sync" button should be deleted (@kassin) as there is no real use case for syncing the color/opacity.

astrofrog:
Just to clarify, in glue there are use cases for the sync button, just not in specviz, so we could selectively delete it in cubeviz.

Issue #342, opened 2018-03-28
Glue slider and wavelength display does not respond to change in units


drdavella:
![image](https://user-images.githubusercontent.com/2458487/38047001-1783b324-328f-11e8-83ae-6ab199196408.png)

![image](https://user-images.githubusercontent.com/2458487/38046986-0cc2010c-328f-11e8-9c7e-fa1d7dc154cb.png)

cc @astrofrog

drdavella:
IMO an acceptable solution would be for cubeviz to hide the glue-specific slider and wavelength display...

kassin:
I agree with @drdavella

SaOgaz:
@drdavella can this issue be closed?

drdavella:
No, this still doesn't work. I will reiterate what I said above, though, which is the easiest/most reasonable solution would be just to hide this slider.

PR #384, opened 2018-04-09
Remove sync button and internally default to False

astrofrog:
There are use cases for the sync button in glue, but it doesn't seem to be necessary in cubeviz, so this PR removes it. If people want to keep the button, I can remove the changes that remove it and keep the part that defaults to False.

**DO NOT MERGE** - this includes https://github.com/spacetelescope/cubeviz/pull/380 and needs to be rebased once https://github.com/spacetelescope/cubeviz/pull/380 is merged

Fixes https://github.com/spacetelescope/cubeviz/issues/326


drdavella:
@astrofrog this looks good to me. I just rebased so it's ready to merge. However, do you think it would make sense to address https://github.com/spacetelescope/cubeviz/issues/342 here as well?

astrofrog:
#342 is going to be harder to address, so I'd like to keep those separate

PR #391, opened 2018-04-12
Added back in the missing yaml files

brechmos-stsci:
The setup_package was missing the yaml files. Added back in.


brechmos-stsci:
Yeah, I think I somehow deleted the '*.yaml' part of the file and so had to add this back in. The travis tests worked and people are probably in bed or don't want to be checking this at almost 11pm, so I am going to force merge and face the consequences.

PR #397, opened 2018-04-13
Moment Maps Tests

javerbukh:
Only testing 1st moment map at the moment


brechmos-stsci:
I don't believe we need this for the 0.2.x branch so will leave off that for now.

PR #403, opened 2018-04-16
Update installation.rst

kassin:
pointed to installation page on wiki

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -1,9 +1,5 @@
-************
-Installation
-************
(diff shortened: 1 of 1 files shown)


brechmos-stsci:
@kassin I am not sure about the changes for these docs.  It looks like you only changed the formatting from RST to Markdown but is that what you intended?

kassin:
Yes.  I want the links to show up properly and had the incorrect syntax for that before... I believe.

PR #409, opened 2018-04-19
Moment Map Tests: changed atol to rtol for more precission

javerbukh:


brechmos-stsci:
Not going to merge into v0.2.x but will merge into master.

PR #411, opened 2018-04-20
Pin glue-core version in CI builds to latest known to work

drdavella:


brechmos-stsci:
Not needed as #413 was merged.

PR #419, opened 2018-05-01
Update to the readme for the 0.2.0 release

brechmos-stsci:
@kassin we can talk about this at 1pm today


brechmos-stsci:
Updated the v0.2.x branch as well.

PR #420, opened 2018-05-02
Dependency updates and contact email change

brechmos-stsci:
The dependency list was missing spectral-cube so I added that in there.


brechmos-stsci:
This will be overtaken by other PRs coming up.

PR #427, opened 2018-05-09
Converted the unicode double quote to normal person double quote

brechmos-stsci:
This appears to be copy paste error that added unicode into the REAMDE file which Windoze was not too happy with.  So, changed it back to regular ascii.


brechmos-stsci:
Merged change into v0.2.x as well.

PR #452, opened 2018-11-26
Version fixes for moving ahed

brechmos-stsci:
This sets the current version to 0.3.dev, fixes an import issue in cubeviz/__init__.py and version dependencies for glue core and specviz

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ env:
 
         # List other runtime dependencies for the package that are available as
         # conda packages here.
-        - CONDA_DEPENDENCIES='pyqt matplotlib scipy glue-core=0.13 asdf specviz=0.5 specutils=0.2.2'
+        - CONDA_DEPENDENCIES='pyqt matplotlib scipy glue-core>=0.13 asdf specviz=0.5 specutils=0.2.2'
         - CONDA_VERSION=4.3.31
 
         # List other runtime dependencies for the package that are available as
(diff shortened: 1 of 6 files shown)


astrofrog:
I'll look into why cubeviz doesn't work with glue 0.14

astrofrog:
Note: I suggested to @brechmos-stsci on Slack to remote matplotlib from CONDA_DEPENDENCIES because I think we are running into an issue where sip is getting pinned to 4.18:

https://github.com/astropy/ci-helpers/blob/4303c636b30f1704dd65521ac37baf0d13b85207/appveyor/install-miniconda.ps1#L401

but glue requires a more recent version.

astrofrog:
I think ``MATPLOTLIB_VERSION`` is ignored on AppVeyor. Try ``matplotlib=2.2`` in ``CONDA_DEPENDENCIES``.

brechmos-stsci:
I have tried a few combinations of astropy, glue, matplotlib and none of them have worked with appveyor using Python==3.5.  @astrofrog  is going to take a look at this.

brechmos-stsci:
There was a discussion about whether we should even test against Python3.5.  It was decided that we would not test against Python 3.5 as condaforge is not building astropy against Python 3.5 and we were pulling an older version. So, I removed Python 3.5 and added Python 3.7.

cc @astrofrog

brechmos-stsci:
So it appears this failed with glue-core 0.13 and works with glue-core 0.14. 

This needs to be merged into master and then merged into 0.2.x and 0.2.1 and then a pypi release on 0.2.1 has to happen

PR #456, opened 2018-12-05
Use travis for Windows build instead of appveyor

drdavella:


.travis.yml
```diff
@@ -65,9 +65,11 @@ matrix:
     fast_finish: true
 
     include:
-        # Try MacOS X
+        # Try MacOS X and Windows
         - os: osx
 
+        - os: windows
```
brechmos-stsci: What version of windows does it run on?  We probably want Windows 10, but do we want WIndows 7 as well?

drdavella: The only Windows version supported by travis right now is [Windows Server 1803](https://docs.travis-ci.com/user/reference/windows/#windows-version). I'm not sure this is a problem, though, since the main point is just to provide a smoke test for all Windows systems.

drdavella:
After merging this I disabled the Appveyor webhook for this repo and also turned off tests in Appveyor. The project is still there, though, so if someone decides to use it again it can be re-enabled.

PR #462, opened 2019-01-23
ReadTheDocs fixes

brechmos-stsci:
There were issues that came up several months ago with the read the docs.  It had to do with astropy changes etc. So, converted the docs from MD to RST. And there were some other fixes that needed to happen including removing the requirement to install cubeviz to get the version etc.


drdavella:
Looks good but we probably should figure out the test failures before proceeding.

brechmos-stsci:
Test are running fine and the read the docs page for this branch appears correct. Going to merge this (fyi 
 @drdavella)

PR #464, opened 2019-01-25
Fix use of pytest fixtures to avoid errors with pytest>=4.1

drdavella:
Calling fixtures directly had been deprecated in `pytest` for a while, but it's no longer supported at all in 4.1. The easiest fix is simply to factor out the logic of the fixture into a standalone function that can be called both by the existing fixture and also separately when needed.

### DIFF
--- a/cubeviz/tools/tests/test_collapse_cube.py
+++ b/cubeviz/tools/tests/test_collapse_cube.py
@@ -16,8 +16,7 @@ from ...tests.helpers import (toggle_viewer, select_viewer, left_click,
 DATA_LABELS = ['018.DATA', '018.NOISE']
 
 
-@pytest.fixture(scope='module')
-def collapse_cube(cubeviz_layout):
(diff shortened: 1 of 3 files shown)


brechmos-stsci:
We had to merge #465 first and then rebase this as the pegged conda version was causing issues in Travis.

brechmos-stsci:
We had a long discussion about some of the fixture issues. It seemed best to just remove the fixture from the last test in the collapse cube test as adding the ROI to the viewer was not working properly. 

The fixture mechanism changed from pytest v3 to v4 and so had to explicitly pass in the fixture rather than reference as a method call in the test methods.

drdavella:
To clarify, the reason the fixture didn't work in that particular test was that the test actually modified the underlying cubeviz application **after** the tool GUI was created by the fixture, so things were out of sync. I don't think this indicates an underlying issue with the application itself.

PR #467, opened 2019-02-04
WIP: Include coverall coverage test

brechmos-stsci:
Going to see about adding the coverall coverage test to cubeviz.


brechmos-stsci:
Going to close this PR as @drdavella is switching the build env stuff over to tox and will include coveralls in it.

PR #474, opened 2019-02-06
Use tox for CI testing

drdavella:
This greatly simplifies the CI config files and also allows test environments to be reproduced locally with minimal effort.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,6 +58,9 @@ matrix:
         - os: windows
           env: TOXENV='py36'
 
+        # Test against Py37 (expected to fail for now)
+        - env: TOXENV='py37'
+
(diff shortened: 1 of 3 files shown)


brechmos-stsci:
Not sure if it is in there, but we should probably have py3.7 in there, maybe as an expected fail (?).

brechmos-stsci:
Looks good. Only thing I would suggest is a short Wiki page on how I can create / run a test environment locally based on the tox file. (please).  Then maybe link here just for reference.

drdavella:
Added some documentation [here](https://github.com/spacetelescope/cubeviz/wiki/How-to-Run-Tests#reproducing-ci-tests-using-tox).

PR #482, opened 2019-02-12
Added basic text of IFU FITS file requirements

### LINKED ISSUE #256: Document requirements for data cubes to load into cubeviz
Write up what the minimum requirements are for loading data into cubeviz.

brechmos-stsci:
@kassin and @hcferguson I believe adding some documentation into the read-the-docs is the best way to document this.  So, I opened this PR up to start a discussion of how detailed you think this should be from a scientist point of view. 

The most basic description is essentially what is in this initial PR but we could make this as long as we wanted as there are lots of caveats etc etc. So.... thoughts on text you think you would like to see here? 

Fixes #256

### DIFF
--- a/docs/readingindata.rst
+++ b/docs/readingindata.rst
@@ -29,20 +29,26 @@ Create an Data Cube
 ===================
 
 CubeViz has several expectations on the format of the FITS file in order to
-easily read in IFU data.  At a bare minimum, the FITS file must have at least
-one data extension that has an extension name. If there is more than one data
-extension that is expected to be read in, then they must all be of the same
-size. If the extension name is not set, then it will be auto-set to be the HDU
-number.
+easily read in IFU data. At a bare minimum, the FITS file must contain an image
+HDU that contains the three-dimensional science data. By convention this
+extension is often named `'SCI'`.
 
-The CTYPE1 should be set to `RA---TAN`, CTYPE2 should be set to `DEC---TAN` and
+CubeViz is able to display multiple components from a single data set. Some
+files might wish to provide error and data quality information, for exaple.
+If the FITS file provides these, each must be stored in a separate image HDU
+These are often named `'ERR'` and `'DQ'` by convention. The data components of
+these HDUs must match those of the HDU containing science data.
+
+Each HDU header must contain several keywords with WCS information. The CTYPE1
+keyword should be set to `RA---TAN`, CTYPE2 should be set to `DEC---TAN` and
 CTYPE3 should be set to one of the valid spectral coordinate types
 `SPECTRAL_COORD_TYPE_CODES` listed at the top of the `reader file
 <https://github.com/spacetelescope/cubeviz/blob/master/cubeviz/data_factories/ifucube.py>`_.
 
-The CUNIT1 and CUNIT2 should be set to `'deg'` and CUNIT3 must be set to a
-valid FITS compatible units (e.g., `m`, `um`, or `AA`). You can check if a unit
-is FITS-compatible by using Astropy's Unit class:
+The CUNIT1 and CUNIT2 keywords should be set to `'deg'` and CUNIT3 must be set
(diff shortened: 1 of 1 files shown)


kassin:
I agree to this.

hcferguson:
Reading through the text, I have a few comments/suggestions.

- Email to Susan Kassin probably isn't appropriate for docs. Can send them to stars.stsci.edu or to Github. 
- Does cubeviz require the first two axes to be spatial and the third to be spectral? Or can that be over-ridden in the yaml file.
- Do the CUNITS really have to be 'deg' or is any FITS-understood angular unit okay (e.g. radians)?
- Are the spectral units being checked by Astropy or by the FITS reader? My impression was that the FITS reader was doing the checking, so the spectral units might need to be FITS compatible, not just Astropy compatible?
- Wording: I wouldn't call the yaml files themselves "data loaders". Would suggestion something like "CubeViz uses yaml files to specify the information needed to load data files." or something like that.

brechmos-stsci:
@hcferguson 
I'll change to stars and github issue suggestion

Currently, as I remember, the first two axes have to be spatial and the third spectral.  We had not done any re-ordering of the axes when reading in.  

For the CUNITS we are setting them to 'deg' if that field is empty or missing from the header. I don't believe we had any IFU data that actually had the CUNIT fields set correctly. I would have to look back at those.

We are comparing the spectral units to a list of valid spectral astropy defined units.

I'll change the yaml file data loader desc.

brechmos-stsci:
@hcferguson One other note on the axes and CUNIT parts, we were aiming at being able to load the data we had available to us, which, at that point all were spatial-spatial-spectral and had typically missing units. 

But... if you know of other IFU data that might be different (radians or spectral-spatial-spatial), happy to add that into the mix for testing and re-working the loader.

docs/readingindata.rst
```diff
@@ -20,6 +20,24 @@ should appear.  In the dialog box, select the data cube you wish to load and
...
+size. If the extension name is not set, then it will be auto-set to be the HDU
+number.
+
+The CTYPE1 should be set to `RA---TAN`, CTYPE2 should be set to `DEC---TAN` and
+CTYPE3 should be set to one of the valid spectral coordinate types
+`SPECTRAL_COORD_TYPE_CODES` listed at the top of the [reader
+file](https://github.com/spacetelescope/cubeviz/blob/master/cubeviz/data_factories/ifucube.py).
+
+The CUNIT1 and CUNIT2 should be set to `'deg'` and CUNIT3 must be set to a
+valid Astropy units spectral_axis compatible unit (e.g., `m`, `um`, or `AA`).
```
astrofrog: I agree with @hcferguson that this should also say 'FITS-compatible' - note for example that ``micron`` is not a valid FITS unit (I saw it in some of the test files). We could always say something like:

*You can check if a unit is FITS-compatible by using Astropy's Unit class:*

```
from astropy.units import Unit
unit = Unit('m', format='fits')
```

*This will crash for units not compatible with the FITS standard.*

brechmos-stsci:
@hcferguson You good with what is here now for the docs?  If so, if you could approve the PR.  @kassin already said she was good with gist of it.

hcferguson:
Looking at the yaml file and reading the docs, my conclusion is that if I have a non-standard file (e.g. with the spectral axis in the first dimension rather than the third, or with wavelength units in furlongs), I'm up a creek. Even a custom loader isn't going to help; I'll have to reformat my data. 

If that's the correct conclusion, then I'm fine with the text. If there is some sneaky way with a yaml file to let me specify weird units or a different organization of the array then we ought to document that.

brechmos-stsci:
@hcferguson That is the correction assumption (for now). The cubeviz loader will not re-organize. When we have some datasets available that are of an other format then we can re-organize things to be able to load them.

hcferguson:
Good to go for this PR.

Issue #301, opened 2018-03-14
Specviz spectral plot display when region selection and multiple flux cubes


brechmos-stsci:
There is a bit of a question of what to show in specviz if we have mulitple flux cubes in cubeviz and a region showing on the cubes.

For example, if we have a FLUX cube and spatially smooth it. Then we draw a region in an image viewer.  Then... I suspect we want to see multiple spectra in specviz corresponding to the region applied to each FLUX or FLUX derived cube.

@kassin @drdavella

kassin:
Is this incorporated in #299 ?

brechmos-stsci:
Not really, #299 is related with data loading, should it be possible to have two copies of the same data loaded etc.  

This ticket is strictly to do with display of spectra from two related FLUX datasets (for example the original FLUX and a smoothed version).

hcferguson:
Need to check what the behavior is once we link to the new SpecViz, and decide if we are happy with that.

SaOgaz:
Reviewed current version, has product owner (@hcferguson) approval for now for current behavior.

PR #484, opened 2019-02-13
Integration of latest SpecViz

### LINKED ISSUE #330: Autorange issue in specviz window with changes to spectral plots
There a couple issues with the auto range of the dispersion and flux axes when the wavelength units are changed, when a new ROI is created (and the median spectrum is hidden).

The first one probably is to do with the green bar not being updated properly with units changes.

The second one I believe is happening is that pyqtgraph (in specviz) does an autoRange (to set the "proper" range of the x and y axes) but it is doing it for all spectra, not just displayed spectra. 

The third one seems to be if I change the redshift value, then the green bar goes awry.

cc @kassin

### LINKED ISSUE #364: Check cubeviz for listener on units change from specviz
According to @nmearl cubeviz is not receiving units change information from specviz https://github.com/spacetelescope/specviz/issues/382#issuecomment-378736431

### LINKED ISSUE #386: Specviz no longer responds to flux units changes
This is being tracked in https://github.com/spacetelescope/specviz/issues/392.

### LINKED ISSUE #477: Incorporate latest Specviz into Cubeviz

nmearl:
Requires spacetelescope/specviz#671.

This resolves #386, and resolves #364, and resolves #330. It also resolves #477.


astrofrog:
I just tried this out and ran into the following:

* With the ``manga-7495-12704-LOGCUBE.fits`` test cube I get:

```
'1E-17 erg/s/cm^2/Ang/spaxel' did not parse as unit: At col 17, Ang is not a valid unit. Did you mean aG, aN, ag, nG or ng?
```

* By default it looks like specviz is occupying too much space:

screenshot 2019-02-20 at 16 45 50

I'm using a 13" screen. I can't actually shrink specviz without making it disappear completely (i.e. the minimum size is too large)

* There is no longer an indicator of the current slice in specviz

* If I click on the vertical line for zero in specviz and drag, a new region is created

I was also asked by @javerbukh to check the status of a few issues:

* https://github.com/spacetelescope/cubeviz/issues/386 (Specviz no longer responds to flux units changes) is still a problem

* https://github.com/spacetelescope/cubeviz/issues/364 (Check cubeviz for listener on units change from specviz) is still a problem

* I can't test https://github.com/spacetelescope/cubeviz/issues/330 because there is no longer an indicator of the current slice in specviz

* https://github.com/spacetelescope/cubeviz/issues/301 is more of a design decision about behavior, so not really something we should worry about here if we care about feature parity.

nmearl:
Thanks @astrofrog!

>     * If I click on the vertical line for zero in specviz and drag, a new region is created

Can you elaborate on this a bit, I'm not sure I follow?

>     * I can't test #330 because there is no longer an indicator of the current slice in specviz

The latest commits should include the slice indicator.

javerbukh:
1. Specviz seems to work as expected with the following files: `DATACUBE_HDFS_v1p0.fits.gz`, `FLAMES_Haro11_L6_cube_bjames.fits`, `GMOS_NGC4449_red_bjamesv2.fits`, `SINFONI_CAS64_k_bjames.fits`, `KLASS_KMOS_COMBINE_SCI_RECONSTRUCTED_S2A_1261.fits`
2. The following file types, data did not appear in the Specviz window because of an error: `JWST-NIRSpec-IFU-GAL2136_CLEAR_PRISM.fits `(error: "Failed to add data
'COUNTS METER-1 SLICERPOS-2' did not parse as unit: At col 0, COUNTS is not a valid unit. Did you mean count?"), `manga-7495-12704-LOGCUBE.fits` (error: same as Tom's), and `bx442_OSIRIS.fits` (error: "Data is not a 1D spectrum").
3. The UI is looking good and I am not getting the same spacing issue that Tom has in his screenshot (I am on a 15 inch macbook).
4. The slice indicator is working as expected.
5. I was not able to replicate what Tom was mentioning about creating a new region from clicking the vertical line above 0.

hcferguson:
I tried jw10001001001_01101_00001_mirifushort_s3d_ref.fits in Nick's master cubeviz. It gives a bunch of warnings in the popup box:
```
Checking filename /Users/hcferguson/datb_data/cubeviz_190222/jw10001001001_01101_00001_mirifushort_s3d_ref.fits

HDU /Users/hcferguson/datb_data/cubeviz_190222/jw10001001001_01101_00001_mirifushort_s3d_ref.fits_0
    CTYPE1 is NONE, setting to RA---TAN
    CTYPE2 is NONE, setting to DEC--TAN
    CTYPE3 is NONE, setting to WAVE-TAB
    CUNIT1 is NONE, setting to deg
    CUNIT2 is NONE, setting to deg
    CUNIT3 is NONE, setting to um
    Setting HDU 0 EXTNAME field to /Users/hcferguson/datb_data/cubeviz_190222/jw10001001001_01101_00001_mirifushort_s3d_ref.fits_0

HDU DQ
    CTYPE1 is NONE, setting to RA---TAN
    CTYPE2 is NONE, setting to DEC--TAN
    CTYPE3 is NONE, setting to WAVE-TAB
    CUNIT1 is NONE, setting to deg
    CUNIT2 is NONE, setting to deg
    CUNIT3 is NONE, setting to um

HDU ERR
    CTYPE1 is NONE, setting to RA---TAN
    CTYPE2 is NONE, setting to DEC--TAN
    CTYPE3 is NONE, setting to WAVE-TAB
    CUNIT1 is NONE, setting to deg
    CUNIT2 is NONE, setting to deg
    CUNIT3 is NONE, setting to um

HDU WMAP
    CTYPE1 is NONE, setting to RA---TAN
    CTYPE2 is NONE, setting to DEC--TAN
    CTYPE3 is NONE, setting to WAVE-TAB
    CUNIT1 is NONE, setting to deg
    CUNIT2 is NONE, setting to deg
    CUNIT3 is NONE, setting to um
```
If I click "accept" I see some data, but the image windows are inactive (can't get the mouse focus away from the specviz window in order to use the menus; but the cursor does show values).

Traceback on the terminal is:
```
(cubeviz-master) ~/d/cubeviz_190222 ❯❯❯ cubeviz
objc[2251]: Class FIFinderSyncExtensionHost is implemented in both /System/Library/PrivateFrameworks/FinderKit.framework/Versions/A/FinderKit (0x7fff966d6cd0) and /System/Library/PrivateFrameworks/FileProvider.framework/OverrideBundles/FinderSyncCollaborationFileProviderOverride.bundle/Contents/MacOS/FinderSyncCollaborationFileProviderOverride (0x13c53acd8). One of the two will be used. Which one is undefined.
WARNING:ifcube: HDU 0 has no EXTNAME field
WARNING:glue.core.coordinates:

*******************************
Encounted an error during WCS parsing. Discarding world coordinates! 
ERROR 5 in wcsset() at line 2176 of file cextern/wcslib/C/wcs.c:
Invalid parameter value.
ERROR 3 in tabset() at line 749 of file cextern/wcslib/C/tab.c:
Invalid tabular parameters: Each element of K must be positive, got 0.

*******************************


/Users/hcferguson/anaconda3/envs/cubeviz-master/lib/python3.6/site-packages/glue/core/data.py:1637: UserWarning: Data.visible_components is deprecated
  warnings.warn('Data.visible_components is deprecated', UserWarning)
Traceback (most recent call last):
  File "/Users/hcferguson/anaconda3/envs/cubeviz-master/lib/python3.6/site-packages/glue/app/qt/application.py", line 236, in _choose_load_data_wizard
    self._choose_load_data(data_importer=data_wizard)
  File "/Users/hcferguson/anaconda3/envs/cubeviz-master/lib/python3.6/site-packages/glue/app/qt/application.py", line 255, in _choose_load_data
    app.add_datasets(app.data_collection, data)
  File "/Users/hcferguson/anaconda3/envs/cubeviz-master/lib/python3.6/site-packages/glue/core/application_base.py", line 302, in add_datasets
    data_collection.extend(datasets)
  File "/Users/hcferguson/anaconda3/envs/cubeviz-master/lib/python3.6/site-packages/glue/core/data_collection.py", line 98, in extend
    self.append(d)
  File "/Users/hcferguson/anaconda3/envs/cubeviz-master/lib/python3.6/site-packages/glue/core/data_collection.py", line 84, in append
    self.hub.broadcast(msg)
  File "/Users/hcferguson/anaconda3/envs/cubeviz-master/lib/python3.6/site-packages/glue/core/hub.py", line 217, in broadcast
    handler(message)
  File "/Users/hcferguson/anaconda3/envs/cubeviz-master/lib/python3.6/site-packages/cubeviz/listener.py", line 51, in handle_new_dataset
    self.configure_layout(data)
  File "/Users/hcferguson/anaconda3/envs/cubeviz-master/lib/python3.6/site-packages/cubeviz/listener.py", line 61, in configure_layout
    self.setup_data(cubeviz_layout, data)
  File "/Users/hcferguson/anaconda3/envs/cubeviz-master/lib/python3.6/site-packages/cubeviz/listener.py", line 120, in setup_data
    cubeviz_layout.add_data(data)
  File "/Users/hcferguson/anaconda3/envs/cubeviz-master/lib/python3.6/site-packages/cubeviz/layout.py", line 637, in add_data
    self._flux_unit_controller.set_data(data)
  File "/Users/hcferguson/anaconda3/envs/cubeviz-master/lib/python3.6/site-packages/cubeviz/controls/flux_units.py", line 326, in set_data
    wcs = data.coords.wcs
AttributeError: 'Coordinates' object has no attribute 'wcs'
```

hcferguson:
Hmm...possible clue: Even though I selected jwst-ifu-fits, it's showing the filename as `manga: jw10001001001_01101_00001_mirifushort_s3d_ref`.

drdavella:
@nmearl it looks like this needs to be rebased. Also there are some real failures on CI that need to be addressed.

nmearl:
Indeed. Most of the specific failures will fall out once the specviz PR is merged.

nmearl:
@drdavella It seems that the tests are failing because of fundamental incompatibilies between glue 0.13 and glue 0.14+. I'm not sure we can refactor the specviz data viewer to be compatible with both. I'm waiting on a response to @astrofrog to see if the schism between the versions is too great, or if it really is maybe just some easy refactoring (my preliminary look over the code base suggests it's not, but Tom might have some guru-level code bridge that could Solve All The Problems™️).

How amenable are we to dropping support for 0.13 if using the latest specviz?

drdavella:
@nmearl I'm perfectly amenable to dropping support for 0.13. We have no compelling reason to support more than one version right now.

cubeviz/conftest.py
```diff
@@ -65,7 +65,7 @@
 from .tests.helpers import (toggle_viewer, select_viewer, create_glue_app,
                             reset_app_state)
 
-@pytest.fixture(scope='session')
+@pytest.fixture(scope='module')
```
drdavella: Is this going to be necessary in the long term or was it just an experiment to try to make the tests pass?

nmearl: I'm not quite sure how to tackle this issue: locally, it seems that a session-scoped fixture with the `reset_state` function is not enough to clear the state of the gui. The only way I could get the tests to pass locally was if the applications was hard reset before every module.

nmearl: I think we'll have to take a look at the `reset_app_state` in greater detail. But the tests were not significantly longer when scoped with `module`.

hcferguson:
@nmearl From my perspective, it's critical to have all of the viz's in sync on which version of glue they require. Actually, it's even important to extend this across all the tools, including ginga, so that we can have one Astroconda distribution rather than several. If some big change to glue comes along to force us to refactor code, we should try to sync that up.

I'm not sure what the implications of this general philosophy are for this particular case.

Looking at the Astroconda packages, it looks like it's several versions of glue back. So this might require consultation with @pllim about ginga/glue dependency and @jhunkeler about Astroconda.

pllim:
Ginga has no dependency on Glue and given that I am not involved in the Viz stuff, I have no strong opinion. Did you consult Tom R on this? You can go the route of pipeline and define a specific pinned environment YAML for your users. :woman_shrugging: 

The last time we did a max version pinning at astroconda, I think it ended in disaster. I wouldn't advise it unless you have no other choice.

nmearl:
I cannot reproduce the error in `test_ui` locally -- @drdavella could you try running the test suite locally and let me know if you're seeing it?

SaOgaz:
I spoke to @jhunkeler offline (well, off github) about the astroconda glue issue, and he made a reasonable case for us not being able to support glue in astroconda.  I think we can still keep our tool releases using consistent versions of glue, while having users pull glue from other sources (from the glue channel or from conda-forge).

drdavella:
@nmearl I'm not seeing the failure locally either. Will try to investigate more in a bit.

drdavella:
@nmearl can you try setting the `pytest` fixture scope back to `"session"`? The tests seem to work fine for me locally when using session scope, and I'm wondering if it will resolve the error we see currently.

drdavella:
See #498 which reverted the fixture scope from `"module"` back to `"fixture"`. This appeared to make the tests pass.

cubeviz/controls/slice.py
```diff
@@ -122,7 +124,7 @@ def change_slider_value(self, amount):
         new_index = self._slice_slider.value() + amount
         self._slice_slider.setValue(new_index)
 
-        specviz_dispatch.changed_dispersion_position.emit(pos=new_index)
+        # specviz_dispatch.changed_dispersion_position.emit(pos=new_index)
```
brechmos-stsci: Is this going to be handled differently now?  If so, can probably delete the commented line.

cubeviz/controls/slice.py
```diff
@@ -2,7 +2,7 @@
 
 from glue.core import HubListener
 from glue.utils.array import format_minimal
-from specviz.third_party.glue.data_viewer import dispatch as specviz_dispatch
+# from specviz.third_party.glue.data_viewer import dispatch as specviz_dispatch
```
brechmos-stsci: There are several places with commented lines that could probably be deleted if they are not going to be used now.

cubeviz/controls/tests/test_slice_controller.py
```diff
@@ -353,10 +353,10 @@ def test_wavelength_ui(qtbot, cubeviz_layout, slice_index=1000):
 
     # Units and negative redshift
     wui.do_calculation(wavelength_redshift=-3, wavelength_units='um')
-    assert_wavelength_text(cubeviz_layout, '-1.1029')
+    assert_wavelength_text(cubeviz_layout, '-9.6250e-01')
```
brechmos-stsci: Why did the value change here?  it is different than the -1.1029 from previously.

nmearl: I can't say for certain, but I believe it's because previously, cubeviz would immediately update its current wavelength location based on the vertical line in specviz. Now, however, the initial position is read directly from cubeviz and used to set the vertical line instead.

drdavella:
After testing this locally I noticed that ROI deletion in cube viewers does not propagate properly the spectrum viewer. The corresponding subset appears to be removed from the list of data items in `specviz`, but the deletion is not reflected in the plot itself.

All other ROI updates appear to propagate correctly, though.

nmearl:
@drdavella Fixed in spacetelescope/specviz#680.

cubeviz/controls/slice.py
brechmos-stsci: If these commented lines aren't need, maybe best to delete them.

nmearl:
[spacetelescope/specviz#680](https://github.com/spacetelescope/specviz/pull/680) has been merged.

nmearl:
@drdavella @brechmos-stsci Any chance of a re-review?

cubeviz/controls/wavelengths.py
```diff
@@ -88,7 +84,9 @@ def update_units(self, units):
         self._send_wavelength_unit_message(units)
         self._send_wavelength_message(self._wavelengths)
 
-        specviz_dispatch.changed_units.emit(x=units)
+        self._cv_layout.specviz._widget.hub.plot_widget.spectral_axis_unit = units
+        self._cv_layout.specviz._widget._slice_indicator.setPos(
+            self._wavelengths[self._cv_layout._active_cube._widget.slice_index])
```
drdavella: Just curious but why are these lines necessary? I would have thought that this would have been handled by the two messages that were sent above.

cubeviz/controls/slice.py
```diff
@@ -193,7 +194,7 @@ def _on_slider_released(self):
 
         index = self._slice_slider.value()
         self._send_index_message(index)
-        specviz_dispatch.changed_dispersion_position.emit(pos=index)
+        # specviz_dispatch.changed_dispersion_position.emit(pos=index)
```
drdavella: Here's another commented line that should be deleted.

setup.cfg
```diff
@@ -22,7 +22,7 @@ install_requires =
     astropy
     asdf
     glue-core>=0.13
```
drdavella: We now require `glue>=0.14` right?

cubeviz/controls/wavelengths.py
nmearl: There's a race condition with the messages sent out from cubeviz due to the bi-directional connection on the slice index indicators. I'll push a workaround, but it's not _great_ (uses signal blockers), but slightly more in line with what you're getting at here.

setup.cfg
nmearl: That's right. Updated.

PR #485, opened 2019-02-14
WIP: Update dependency requirements and min python version

### LINKED ISSUE #470: v0.3 release dependencies
As part of the v0.3 release and after some discussion about package dependencies with @drdavella we will need to come up with a set of minimum versions (or ranges) for cubeviz dependencies.  Then the following files will have to be updated based on the decisions. 

- [ ] Update setup.cfg
- [ ] Update .travis.yml
- [ ] Update astroconda cubeviz build recipe

Packages and versions to consider:

* Python = 3.6 or Python = 3.7
 * astropy ( ver ?)
 * glue-core 0.13.x or 0.14.x (?)
 * numpy (>= 1.13) ?
 * specviz 0.6.x (?)
 * spectral-cube >=0.4.2 (?)
 * matplotlib
 * asdf

The above dependencies will need to be checked with Joe to make sure they are available in the astroconda release, too.  Once they are decided then the above files will need to be updated.

brechmos-stsci:
This PR updates the minimum versions etc for the dependencies and also sets the minimum Python version to 3.6. 

See discussion and notes in #470 (based on a discussion with @drdavella).

Fixes #470

### DIFF
--- a/cubeviz/__init__.py
+++ b/cubeviz/__init__.py
@@ -4,6 +4,8 @@
 This is an Astropy affiliated package.
 """
 
+__minimum_python_version__ = "3.6"
+
 # Affiliated packages may add whatever they like to this file, but
 # should keep this content at the top.
 # ----------------------------------------------------------------------------


drdavella:
This PR should add tests against Py37 before being merged. This will only be possible once the new `specviz` is fully integrated.

Issue #227, opened 2018-02-16
Add tests for data factories


astrofrog:
As discussed during this sprint, we need tests of the data factories in cubeviz. Rather than including full files in the repository, I suggested on one of the calls taking existing files and keeping only a subset of channels, but leaving everything else untouched - this should make it possible to have FITS files that are only a few hundred kB at most.

brechmos-stsci:
@drdavella We should talk about how to store IFU FITS files for testing.  We did talk a little but should come up with a plan, even in the interim so I can add some tests for some basic cubes.

brechmos-stsci:
@astrofrog and I talked. I am going to send him some of the files, he is going to create cutouts but with the same HDU header formats so that we can use these for testing.

There are two datasets I am checking on to confirm they can be made "public".

brechmos-stsci:
@astrofrog , if you want to include the manga data, first, in a PR just to get it up here, go for it. The manga data is publicly available (https://www.sdss.org/dr15/manga/marvin/). Then we can add to the PR when we know about other data we can use. (Or we can wait, a few days, no prob)

brechmos-stsci:
I talked with @bethanjames  and she is fine with us using the SINFONI data for data loader testing.

PR #486, opened 2019-02-14
Added tests for data factories

astrofrog:
This adds initial tests for the data factories. It's important not to include the full datasets in the repository as they are quite large, so this includes fake small arrays in the data instead of the original arrays, but otherwise preserves all the metadata. If we need larger data for some other tests, then we should use remote data (like astropy does with data.astropy.org).

All datasets included here (MIRI, SINFONI, MANGA) have been cleared to be public (even though we actually replace the data with small fake data anyway). I can include further datasets as they are authorized.

At the moment the test I've included is pretty rudimentary, as it just checks that the data loads with the correct factory, and that the data shape is correct. We could envisage including more sophisticated tests, but it's important to at least make sure most of the generic data factory code runs during the tests.

This also includes a bug fix that came to light with the tests (de752430b63fcb0a6bab1ad027204f382ed0fba2) and I've also removed the config.py file at the root of the repo (not sure why it's there?)

Fixes https://github.com/spacetelescope/cubeviz/issues/227


config.py
```diff
@@ -1,6 +0,0 @@
-# Licensed under a 3-clause BSD style license - see LICENSE.rst
-
-import cubeviz.data_factories
-from cubeviz import layout
-from cubeviz.listener import CubevizManager
-from cubeviz import startup
```
brechmos-stsci: Just out of curiosity, why were these removed?  Is it a change in glue?  I don't remember why I/we added them but I thought they needed to be in here.

brechmos-stsci: Oh, just saw the question as to why it was there. :).  NVM.  Not 100% sure, I don't remember

astrofrog: config.py files are normally user configuration files - they shouldn't be in a package and aren't needed to run cubeviz.

PR #491, opened 2019-02-21
Pin spectral cube version to avoid bug in 0.4.4

drdavella:
This will hopefully serve as a workaround for the errors that have been happening on our nightly cron builds.


drdavella:
This is necessary due to an issue in `spectral_cube` which has been reported: https://github.com/radio-astro-tools/spectral-cube/issues/533

We should merge this workaround for now but will need to update our dependencies once the issue is fixed and released upstream.

PR #492, opened 2019-02-21
Overhaul package infrastructure to use modern practices

drdavella:
~We still rely on `astropy_helpers` to generate a `version.py` file, but instead of including it as a submodule we now put it in `setup_requires` and in `pyproject.toml`.~


setup.cfg
```diff
@@ -1,3 +1,36 @@
...
+long_description = file: README.md
+url = https://github.com/spacetelescope/cubeviz
+github_project = spacetelescope/cubeviz
+
+[options]
+packages = find:
+zip_safe = False
+python_requires = >=3.6
+include_package_data = True
+setup_requires = setuptools_scm
```
astrofrog: Btw if you are using pyproject.toml I think you can define it there instead?

drdavella:
I'm going to argue that the coverage report is not particularly relevant for this PR.

brechmos-stsci:
Agreed that the code coverage failures aren't relevant for this PR.

PR #494, opened 2019-02-25
Version import fixes

astrofrog:
The version import doesn't work anymore now that setuptools_scm is used. Also removed a glue file which appeared.

### DIFF
--- a/cubeviz/conftest.py
+++ b/cubeviz/conftest.py
@@ -50,7 +50,7 @@ import os
 # This is to figure out the package version, rather than
 # using Astropy's
 try:
-    from .version import version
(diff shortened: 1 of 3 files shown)


astrofrog:
I'm a little puzzled as to how the tests passed before, but I don't have time to investigate today.

drdavella:
Just putting a comment here for posterity: we don't actually call `cubeviz.main` to start `cubeviz` in the test suite. Instead, we use a test fixture that constructs a `glue` application in a somewhat different way (since we don't need to process command-line arguments or show the splash screen.

We should probably refactor this so that the tests share more code with the way users run the application.

PR #495, opened 2019-02-26
Refactor app initialization to enable better testing

drdavella:
These changes allow the test fixture to overlap more with the code path executed by users when running the application from the command line. This would have allowed the bug fixed in #494 to have been caught by automated tests.


drdavella:
Since this is a refactoring without any new functionality or bug fixes I'm going to argue that the coverage results here are not particularly relevant.

PR #496, opened 2019-02-28
Attempt to fix Windows builds under tox

drdavella:
Attempting to install both `numpy` and `astropy` using `conda` to see if this resolves the issue with Windows builds.


drdavella:
Closing since this appears to have magically worked itself out on its own.

PR #497, opened 2019-02-28
Remove redundant dependencies from tox config

drdavella:
Removing dependencies that should be pulled in either directly or indirectly in during setup.


drdavella:
Closing since everything here was taken care of in #516.

PR #516, opened 2019-03-05
Updates to make cubeviz pip-installable

drdavella:
This adds `pyqt5` to `install_requires` so it does not need to be installed by `conda`. It also addresses the issue that occurs on some OSX machines where `matplotlib` complains about Python not being installed as a framework.


setup.cfg
```diff
@@ -17,6 +17,7 @@ python_requires = >=3.6
 include_package_data = True
 setup_requires = setuptools_scm
 install_requires =
+    pyqt5<5.12
```
drdavella: According to @astrofrog the `5.12` wheels have issues, so we are avoiding them. Maybe eventually this should be `!=5.12`?

drdavella:
Needed to make one update to get the (allowed-to-fail) Windows test passing. But now it works and since it's been reviewed and tests are passing, I'm merging.

PR #525, opened 2019-03-06
Add units to moment map output

brechmos-stsci:


cubeviz/controls/flux_units.py
```diff
@@ -327,6 +333,22 @@ def set_data(self, data):
         if wcs is not None:
             self.wcs = wcs
 
+
+    def add_data(self, data):
```
drdavella: It looks like this method isn't used anywhere. Is it necessary?

brechmos-stsci: twas unnecessary, removed.

PR #527, opened 2019-03-06
WIP: Stat box tests

### LINKED ISSUE #513: Image Viewer tests: Stat box tests
Part of #394

javerbukh:
There are two tests created, one which tests the stat box for the full spectrum and another that tests after an ROI has been created. These are similar to the currently existing tests in Specviz. If there is time, it would be good to add tests for unit conversion.

One question though, is this the best location for this file? Should there be another directory created for testing Specviz integration?

Fixes #513

### DIFF
--- a/cubeviz/tests/test_stat_box.py
+++ b/cubeviz/tests/test_stat_box.py
@@ -1,0 +1,51 @@
+import numpy as np
+from specutils import SpectralRegion, Spectrum1D
+from specutils.analysis import centroid, equivalent_width, fwhm, line_flux
+from specutils.manipulation import extract_region
+
+
+def test_stat_box_full_spectrum(cubeviz_layout):
+    # Pull out stats dictionary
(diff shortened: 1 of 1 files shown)


drdavella:
I think it makes sense to put all specviz-related tests in a file called `test_specviz_integration.py` or something like that.

However, does this test actually exercise any `cubeviz` functionality? If it's purely testing `specviz` features then the test should go in `specviz`. For `cubeviz` we really only want to test the interface between `specviz` and `cubeviz`, i.e. where the two interact.

I think the intent behind this might have actually been to test the statistics summary associated with the image viewers? But I might be wrong.

image

javerbukh:
That makes more sense. @robelgeda can you confirm that that was the original intention?

drdavella:
Given the intent of these tests, we probably should create a file called `cubeviz/tests/test_image_viewer.py` and put most/all of them there, as appropriate.

javerbukh:
Sounds good. I'm going to close this PR then and start working on that.

PR #529, opened 2019-03-06
Stats box tests in image viewer

### LINKED ISSUE #513: Image Viewer tests: Stat box tests

javerbukh:
The current test checks to see that the data presented in the image viewer stats area is parsed correctly. However, the original code uses many different parameters and functions in order to work, so it was difficult to write tests without using some of the image_viewer functions. Any recommended changes or improvements are welcome.

Resolves #513

### DIFF
--- a/cubeviz/tests/test_image_viewer.py
+++ b/cubeviz/tests/test_image_viewer.py
@@ -1,6 +1,25 @@
 import numpy as np
 from glue.core import roi
 
+def test_stats_box_without_subset(cubeviz_layout):
+    """
+    Tests the stat box underneath the ImageViewer when it is the full spectrum
+    """
+    cl_viewer = cubeviz_layout.split_views[1]._widget
+
+    cl_viewer._subset = None
+
+    data = cl_viewer._data[0][cl_viewer.current_component_id][cl_viewer._slice_index]
+
+    wave = cl_viewer.cubeviz_layout.get_wavelength(cl_viewer.slice_index)
+    data_wave = cl_viewer.cubeviz_unit.convert_value(data, wave=wave)
+
+    assert data_wave is not None
+
+    results = (np.nanmin(data_wave), np.nanmax(data_wave), np.median(data_wave), data_wave.mean(), data_wave.std())
+    results_string = r"min={:.4}, max={:.4}, median={:.4}, μ={:.4}, σ={:.4}".format(*results)
+
+    assert results_string == cl_viewer.parent().stats_text.text()
 
 def test_stats_box_with_subset(cubeviz_layout):
     """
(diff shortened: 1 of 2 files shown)


cubeviz/tests/test_image_viewer.py
```diff
@@ -0,0 +1,18 @@
+import numpy as np
+
+def test_stats_box(cubeviz_layout):
+    cl_viewer = cubeviz_layout.split_views[1]._widget
+
+    if cl_viewer._subset is not None:
```
drdavella: What is the purpose of the `if`/`else` here? This test only runs once and the state of the application should always be the same when the test starts, so only one of these branches is ever going to be executed.

javerbukh:
```
    def update_stats(self):

        if self._subset is not None:
            self.show_roi_stats(self.current_component_id, self._subset)
        else:
            self.show_slice_stats()
```
This is the code block from image_viewer.py, which uses that same if / else to determine which version of `show_stats` to use. I decided to mimic that in case that test data set changes, but I can remove it if that makes more sense.

drdavella: I think what we actually want to do is to test both of those code paths in `update_stats` in separate tests, and not have any branching logic in either of the tests. This will require creating (and deleting) a subset in one of the tests. There should be an example of that in at least one of the other tests.

brechmos-stsci:
@javerbukh There are a lot of levels of what could be compared here. 

I think it might be better to create the ROI and grab the text of the number at the bottom of the viewer. Then compare that text, represented as a floating point number, to the actual expected number.  

This way we are testing the actual number people will use to what we expect. It is that number which they might use for science or for interpreting science, so I think we should check *it*.  Does that make sense?

cubeviz/tests/test_image_viewer.py
```diff
@@ -0,0 +1,46 @@
...
+    mask = cl_viewer._subset.to_mask()[cl_viewer._slice_index]
+    data = cl_viewer._data[0][cl_viewer.current_component_id][cl_viewer._slice_index][mask]
+
+    wave = cl_viewer.cubeviz_layout.get_wavelength(cl_viewer.slice_index)
+    data_wave = cl_viewer.cubeviz_unit.convert_value(data, wave=wave)
+
+    assert data_wave is not None
+
+    results = (np.nanmin(data_wave), np.nanmax(data_wave), np.median(data_wave), data_wave.mean(), data_wave.std())
+    assert results == cl_viewer.show_roi_stats(cl_viewer.current_component_id, cl_viewer._subset)
```
drdavella: This is definitely much better. However, I agree with @brechmos-stsci that we might need to actually compare the text that is displayed in the actual widget. I realize this is kind of a pain but I don't think it should be too difficult given what you have here already.

javerbukh: Got it, I'll work on that.

brechmos-stsci:
I like this test.  Funny, I would have tried parsing the string to get the numbers, and then compare the numbers etc etc, this is so much simpler and better. :)

PR #544, opened 2019-03-08
Overlay unit tests

### LINKED ISSUE #514: Image Viewer tests: Overlay testing
Part of #394

javerbukh:
Overlay tests created, however they are interfering with tests from test_ui.py. test_ui.py continuously checks whether the combo.count() equals a certain value, and by creating a moment map and changing overlays, my tests interfere with that count. Looking for how to proceed at this point.

Resolves #514

### DIFF
--- a/cubeviz/tests/test_image_viewer.py
+++ b/cubeviz/tests/test_image_viewer.py
@@ -1,5 +1,18 @@
+import pytest
 import numpy as np
 from glue.core import roi
+
+from cubeviz.tools.moment_maps import MomentMapsGUI
+
+
+@pytest.fixture(scope='module')
+def moment_maps_gui(cubeviz_layout):
+    cl = cubeviz_layout
+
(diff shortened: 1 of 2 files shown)


cubeviz/tests/test_image_viewer.py
```diff
@@ -44,3 +58,48 @@ def test_stats_box_with_subset(cubeviz_layout):
...
+    cubeviz_layout._overlay_controller._alpha_slider.setValue(50)
+
+    assert len(cl_viewer.axes.images) == 2
+
+    # Return to "No Overlay"
+    cubeviz_layout._overlay_controller._overlay_image_combo.setCurrentIndex(0)
+
+    assert len(cl_viewer.axes.images) == 1
+
+    # Remove the moment map data set
```
drdavella: It actually might be a good idea to move all of this logic to the fixture that resets app state after each test.

PR #545, opened 2019-03-08
Image viewer contour tests

### LINKED ISSUE #509: Image Viewer tests: Contour tests
settings, preview, and other components all need units tests.

Part of #394

javerbukh:
Resolves #509


drdavella:
@javerbukh this needs to be rebased now.

drdavella:
According to @brechmos-stsci, calling `axes.get_children()` on the underlying viewer will show a list that contains `matplotlib.collection.LineCollection` objects when the contour is displayed. These objects should not be present when the contour is not displayed, although there will be other children.

Issue #551, opened 2019-03-11
Glue custom slice -- viewer options don't work


hcferguson:
There is a glue feature that is currently enabled via a little 3D icon with a drawing through it. 
<img width="27" alt="image" src="https://user-images.githubusercontent.com/2261329/54146038-834e9000-4405-11e9-9915-f8b81afdb171.png">

That creates a custom "slice view", which has position along the slice on one axis and the spectra along the other axis. This could be handy (and is certainly slick). I didn't understand the feature from the icon, so I'll add that to the docs; I suggest we keep it (not disable it, as discussed in today's tagup).

<img width="648" alt="image" src="https://user-images.githubusercontent.com/2261329/54146087-9cefd780-4405-11e9-9504-ecc270ebd92a.png">

But there does seem to be a bug: The top-left graphics controls seem to do nothing.
<img width="77" alt="image" src="https://user-images.githubusercontent.com/2261329/54146179-c9a3ef00-4405-11e9-8df4-5aa091179aae.png">

hcferguson:
I lied...the contrast stretching (left button) works if you hold down the control key while panning the mouse. Not sure offhand what the righthand control is supposed to do.

The glueviz docs show the slice viewer with a fuller set of controls...including dropdowns for the two that show up for the cubeviz version.
<img width="403" alt="image" src="https://user-images.githubusercontent.com/2261329/54146477-6d8d9a80-4406-11e9-9f23-7203be5f69f0.png">

Issue #547, opened 2019-03-11
Specviz in a separate tab


hcferguson:
One can select a subset from the `Subsets` menu and then select the specviz viewer. The viewer will pop up in a separate tab. When I tried it after having already created a profile viewer in a separate tab, the specviz viewer opened in that tab. The pan controls worked but zoom didn't.

I closed the tab and tried it again with just the specviz viewer and the controls all worked.

nmearl:
It sounds like this would solve itself if we remove the option to create the Glue profile views.

PR #553, opened 2019-03-12
Prune toolbar options in image viewer

nmearl:
Removes the superfluous 1D profile view from the image viewer toolbar. Kind of addresses #547, ~#551~ (we're keeping that option now).

Screen Shot 2019-03-12 at 11 14 48 AM


hcferguson:
The 3D cube slicer actually seems to mostly work and is useful if we could get it to entirely work. See #551.

nmearl:
Yes, the 3D slice extractor is being kept.

PR #556, opened 2019-03-13
Add integration tests for specviz

### LINKED ISSUE #508: Add automated tests for specviz integration
We need to add automated tests for the new `specviz` integration that was introduced in #484. These tests should exercise fairly high-level functionality that relies on integration between `cubeviz` and `specviz`. The goal is to be able to detect breaking changes in either `cubeviz` or `specviz` that cause the integration to no longer work.

Things that need to be tested at minimum (other items are possible as well):
* slice slider navigation in both directions (`cubeviz` updates `specviz` and vice versa)
* unit change propagation in both directions (both flux and wavelength units)

drdavella:
This resolves #508. It is not currently possible to test flux unit changes since the canned test data we use does not appear to have flux units.


cubeviz/tests/test_specviz_integration.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def tests_wavelength_units_update(cubeviz_layout):
+    """
+    Make sure wavelength unit changes in specviz are reflected in cubeviz.
+    """
+    specviz = cubeviz_layout.specviz._widget
+    current_units = str(cubeviz_layout._wavelength_controller.current_units)
+
+    specviz.hub.plot_widget.set_spectral_axis_unit('Angstrom')
```
nmearl: Small change, but you may want to use the [`update_units`](https://github.com/spacetelescope/specviz/blob/master/specviz/third_party/glue/viewer.py#L423) method in the data viewer instead of settings the unit explicitly on the plot widget (since CubeViz uses it this way). Eventually, I want to fix the wavelength update message and have the slice indicator updated in this method as well.

drdavella: @nmearl the reason I didn't do that is because it would require me to pass new values for both spectral units and flux units and I wanted to keep the test as isolated/direct as possible.

drdavella:
**UPDATE**: I've made it possible to test flux units as well and will be pushing more changes shortly.

cubeviz/tests/test_specviz_integration.py
nmearl: @drdavella I'm not sure I follow. You can just call it like `specviz.update_units(spectral_axis_unit='Angstrom')`, the default value of `None` for flux will ensure that the flux units are left alone. Furthermore, the `set_spectral_axis_unit` will raise the `unit_changed` event (this is different from setting the `spectral_axis_unit` directly, which does not raise an event), which is connected back to cubeviz (it shouldn't _do_ anything, but you're basically doing a loop of `cubeviz unit changed -> specviz unit changed -> cubeviz unit changed`).

drdavella: @nmearl sorry, I think I must have been looking in the wrong place or at the wrong function. I'll make the update.

drdavella: It turns out I actually need to call these methods directly rather than the higher-level one. I need the events to occur since the purpose of the test is to determine whether unit change events in specviz propagate properly to cubeviz.

PR #560, opened 2019-03-14
Do not check primary HDU for valid CUNIT/CTYPE

### LINKED ISSUE #559: Invalid/irrelevant header value issues message
When loading the test data cube that is packaged with cubeviz, I get the following message about header value issues:

```
Checking filename cubeviz/tests/data/data_cube.fits.gz

HDU cubeviz/tests/data/data_cube.fits.gz_0
    CTYPE1 is NONE, setting to RA---TAN
    CTYPE2 is NONE, setting to DEC--TAN
    CTYPE3 is NONE, setting to WAVE
    CUNIT1 is NONE, setting to deg
    CUNIT2 is NONE, setting to deg
    CUNIT3 is NONE, setting to um
    Setting HDU 0 EXTNAME field to cubeviz/tests/data/data_cube.fits.gz_0
```

I don't think this makes any sense. Why does the primary HDU need any of these keywords? All of these keywords are present in the two HDUs that contain the actual data, so there doesn't seem to be a problem with this file.

drdavella:
This resolves #559.

To clarify: we do not check the primary HDU for valid CUNIT/CTYPE **unless** the primary HDU contains a data array with three dimensions.

### DIFF
--- a/cubeviz/data_factories/ifucube.py
+++ b/cubeviz/data_factories/ifucube.py
@@ -180,7 +180,7 @@ class IFUCube(object):
         log.debug("Correct value(s) to be used: {}".format(correct))
 
         for ii, hdu in enumerate(self._fits):
-            if ii == 0 or (hasattr(hdu, 'data') and hdu.data is not None and len(hdu.data.shape) == 3):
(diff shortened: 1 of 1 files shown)


drdavella:
To test whether it works, compare opening `cubeviz/tests/data/data_cube.fits.gz` both with and without this change.

PR #566, opened 2019-03-19
updated documentation

hcferguson:
Documentation updates. 

@dravella -- I don't think inter sphinx is working.


hcferguson:
@drdavella I don't think the inter sphinx links are working.

drdavella:
@hcferguson okay I'll take a look.

drdavella:
If I understand correctly, I think that `intersphinx` is intended more for linking to specific API documentation in other packages. So, for example, if we wanted to put a reference to the documentation of a class or function from `specviz` or `glue`, `intersphinx` would make that fairly simple.

It looks like the main intent here is just to link to the top-level documentation of those packages, in which case I would suggest just including the links inline in the documentation itself (see the top of `docs/index.rst` for examples).

drdavella:
Also worth mentioning that the docbuild failure [on travis](https://travis-ci.org/spacetelescope/cubeviz/jobs/508390908#L758) looks meaningful:
```
Warning, treated as error:
/home/travis/build/spacetelescope/cubeviz/docs/analysis.rst:document isn't included in any toctree
```

hcferguson:
I'll push up a couple fixes...

I  think I link to the main doc page in  in all but one case. I tried to link to a spectral-cube subsection on  moments somewhere.

PR #578, opened 2019-03-25
had to pip install for rtd, so sphinx can grab the version number

SaOgaz:
Fixes RTD, or at least it did in the rtd_testing build


SaOgaz:
Use #579 instead.

PR #579, opened 2019-03-25
had to pip install for rtd, so sphinx can grab the version number

SaOgaz:
Use this PR instead...


brechmos-stsci:
This, as I understand, is required as the version number was created by astropy_helpers previously and so no install of cubeviz was required. Since the removal of astropy_helpers, now the version number is created during the installation procedure. So, it now needs to be installed so the docs can get the correct version number.

PR #583, opened 2019-03-29
updated installation instructions

brechmos-stsci:
Docs referred to glueviz installation, need to fix.


brechmos-stsci:
Going to do off of v0.3, not here

PR #597, opened 2021-04-15
Builds "last working release" for Qt-based CubeViz

ibusko:
This PR fixes the old Qt-based CubeViz desktop app to the point it pops up the UI after a local installation from sources in a clean environment.

This doesn't mean the app works. Besides, in order to get it to render the UI at startup, the python version had to be pinned at 3.6.13, so all tests with other python versions are expected to fail. 

Should we care about CI errors for this "last release"?

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -13,7 +13,7 @@ github_project = spacetelescope/cubeviz
 [options]
 packages = find:
 zip_safe = False
-python_requires = ==3.6
+python_requires = ==3.6.13
 include_package_data = True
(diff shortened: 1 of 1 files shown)


rosteen (approved):
I confirmed that after installing in a clean python 3.6.13 environment, the app loads for me (and I can even load our usual testing cube, mostly). I don't think we care about CI errors, although @eteq is probably the one to confirm that.

eteq:
I agree the CI errors aren't worth worrying about!

eteq (approved):

